PR #13, opened 2019-11-01
support case class <--> GenericRecord derivation

ClaireMcGinty:


avro/src/test/scala/magnolify/avro/test/AvroTypeSpec.scala
```diff
@@ -0,0 +1,91 @@
...
+  ): Unit = {
+    ensureSerializable(tpe)
+    val converter = GenericRecordType[T]
+
+    property(className[T]) = Prop.forAll { caseClass: T =>
+      val avroRepr = converter.to(caseClass)
+      val avroCopy = roundtripAvro(avroRepr)
+      val copy = converter.from(avroCopy)
+
+      Prop.all(eq.eqv(caseClass, copy))
```
ClaireMcGinty: I originally included a property test comparing `avroRepr` and `avroCopy` using the default GenericRecord `equals()`, and it failed on Map types when the ordering of map entries was different. a follow up project to this PR could be adding an Equality derivation for Avro types?

PR #22, opened 2019-11-21
add cats Hash derivation

nevillelyh:


cats/src/main/scala/magnolify/cats/auto/CatsMacros.scala
```diff
@@ -49,18 +55,24 @@ private object CatsMacros {
 }
 
 trait LowPriorityImplicits extends LowPriorityGenGroup {
-  implicit def genEq[T]: Eq[T] = macro CatsMacros.genEqMacro[T]
```
anish749: why is this implicit no longer there?

nevillelyh: I movied it further down to below `Hash` since `Hash` extends `Eq`.

cats/src/test/scala/magnolify/cats/test/HashDerivationSpec.scala
```diff
@@ -0,0 +1,47 @@
...
+  private def test[T: Arbitrary: ClassTag: Cogen: Hash](exclusions: String*): Unit = {
+    ensureSerializable(implicitly[Hash[T]])
+    val props = HashTests[T].hash.props.filter(kv => !exclusions.contains(kv._1))
+    for ((n, p) <- props) {
+      property(s"${className[T]}.$n") = p
+    }
+  }
+
+  // Long.## != Long.hashCode
+  test[Integers]("same as universal hash")
```
anish749: The test looks really neat 👍

cats/src/main/scala/magnolify/cats/auto/CatsMacros.scala
```diff
@@ -29,6 +29,12 @@ private object CatsMacros {
     q"""_root_.magnolify.cats.semiauto.EqDerivation.apply[$wtt]"""
   }
 
+  def genHashMacro[T: c.WeakTypeTag](c: whitebox.Context): c.Tree = {
```
anish749: I am unaware of this, but why do we need this redirection? We aren't cleaning up annotations or anything like that, right?

cats/src/main/scala/magnolify/cats/auto/CatsMacros.scala
anish749: Ohh, I see now, thanks.

cats/src/main/scala/magnolify/cats/auto/CatsMacros.scala
nevillelyh: Because the `implicit def genHash[T]: Hash[T]` etc. in `magnolify.cats.auto._` need to pass on `WeakTypeTag` and only macros can do that.

PR #26, opened 2019-11-25
Use unencoded chars funnel for CharSequence

anish749:
I think it would be nice to not force UTF-8 for `CharSequence`

### DIFF
--- a/guava/src/main/scala/magnolify/guava/semiauto/FunnelDerivation.scala
+++ b/guava/src/main/scala/magnolify/guava/semiauto/FunnelDerivation.scala
@@ -60,7 +60,7 @@ trait FunnelImplicits {
   implicit val intFunnel: Funnel[Int] = Funnels.integerFunnel().asInstanceOf[Funnel[Int]]
   implicit val longFunnel: Funnel[Long] = Funnels.longFunnel().asInstanceOf[Funnel[Long]]
   implicit val bytesFunnel: Funnel[Array[Byte]] = Funnels.byteArrayFunnel()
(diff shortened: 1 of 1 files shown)


guava/src/main/scala/magnolify/guava/semiauto/FunnelDerivation.scala
```diff
@@ -60,7 +60,7 @@ trait FunnelImplicits {
   implicit val intFunnel: Funnel[Int] = Funnels.integerFunnel().asInstanceOf[Funnel[Int]]
   implicit val longFunnel: Funnel[Long] = Funnels.longFunnel().asInstanceOf[Funnel[Long]]
   implicit val bytesFunnel: Funnel[Array[Byte]] = Funnels.byteArrayFunnel()
-  implicit val charSequenceFunnel: Funnel[CharSequence] = Funnels.stringFunnel(Charsets.UTF_8)
+  implicit val charSequenceFunnel: Funnel[CharSequence] = Funnels.unencodedCharsFunnel()
```
nevillelyh: What do you think about `stringFunnel`? Should we not force an encoding and use `unencodedCharsFunnel` also? We're not funneling it to disk so no encoding is fine?

anish749: I think Utf8 for String is a reasonable default. Since this is an implicit, the user can choose to override and specify a `Charset` if they choose to for Strings easily.
However for CharSequence we shouldn't enforce an encoding.
The hasher in guava has special handling for utf8 strings (only)

PR #43, opened 2020-01-07
Add more detail to README

anne-decusatis:
Update examples to be a bit more detailed and have comments explaining what they do, clarify what is/isn't included in the derivations

### DIFF
--- a/README.md
+++ b/README.md
@@ -82,14 +82,17 @@ import magnolify.guava.semiauto._
 FunnelDerivation[MyCaseClass]
 ```
 
-Case class type conversion must be called explicitly, specifying the case class used for conversion, rather than deriving entirely with implicits.
+Conversion between a case class instance and an instance of another type (e.g. GenericRecord) containing the same data must be called explicitly/semi-automatically, specifying the case class to be used for conversion, rather than deriving entirely with implicits.
 
 ```scala
+
 // an annotated example with Avro GenericRecord
 import magnolify.avro._
 import org.apache.avro.generic.GenericRecord
+case class MyCaseClass(field: NestedCaseClass) 
+case class NestedCaseClass(innerField: String) // works with other non-String elements, this is just an example
 val typeclass = AvroType[MyCaseClass] // AvroType typeclass instance defines to and from conversions between Case Class and GenericRecord
-val genericRecord: GenericRecord = typeclass.to(MyCaseClass(...)) // instantiate a case class, then pass in
+val genericRecord: GenericRecord = typeclass.to(MyCaseClass(NestedCaseClass("hi"))) // instantiate a case class, then pass in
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -54,42 +82,43 @@ import magnolify.guava.semiauto._
...
 
 ```scala
+// an annotated example with Avro GenericRecord
 import magnolify.avro._
-val t = AvroType[MyCaseClass]
-val r = t.to(MyCaseClass(...))
-val c = t.from(t)
+import org.apache.avro.generic.GenericRecord
+val typeclass = AvroType[MyCaseClass] // AvroType typeclass instance defines to and from conversions between Case Class and GenericRecord
+val genericRecord: GenericRecord = typeclass.to(MyCaseClass(...)) // instantiate a case class, then pass in
```
danielblazevski: Like the expanded docs here, helped me understand this part more. 

I think it wouldn't hurt to make it more explicit, e.g. define 
```scala
case class MyCaseClass(foo: String, bar: Int)=
```
and then pass in actual args here 
```scals
val genericRecord: GenericRecord = typeclass.to(MyCaseClass("hello", 1)) // instantiate a case class, then pass in
```

README.md
```diff
@@ -54,42 +82,43 @@ import magnolify.guava.semiauto._
...
 
-```scala
+// The other three converters work similarly to Avro: 
+
+// an example with BigQuery TableRow
 import magnolify.bigquery._
-val t = TableRowType[MyCaseClass]
-val r = t.to(MyCaseClass(...))
-val c = t.from(t)
+val typeclass = TableRowType[MyCaseClass]
```
nevillelyh: It's using `t` in some places & `typeclass` here. Make them consistent? Doesn't matter which way.

I was being lazy and using `t`, `r`, `c` for type class, record (Avro, BQ, etc.) & case class.

README.md
```diff
@@ -54,42 +82,43 @@ import magnolify.guava.semiauto._
...
 val t = EntityType[MyCaseClass]
 val r = t.to(MyCaseClass(...))
-val c = t.from(t)
+val c = t.from(r)
 
 implicit val uriField = EntityField.from[String](URI.create)(_.toString) // custom field type
-```
 
-```scala
+// an example with TensorFlow TFExample
```
nevillelyh:
```suggestion
// an example with TensorFlow Example
```

README.md
anne-decusatis: I started off trying to convert the names to all longer/more meaningful, but missed some towards the end. I will fix to use longer names throughout

nevillelyh:
Another minor nit: GH CSS doesn't line wrap comments in code boxes. It'd be nice to manually wrap them so you don't have to scroll horizontally when reading.

PR #51, opened 2020-02-06
Protobuf support

### LINKED ISSUE #5: Add `ProtobufType`
For arbitrary protobuf `Message` types. Should be doable by manipulating the field descriptor, etc.

anne-decusatis:
fixes #5 

Slightly different approach than Avro, etc, as the generated protobuf code doesn't allow you to construct Descriptors.

Options and Maps not supported right now, please give feedback on whether it's OK to merge as is. I've tried a few things for maps over the past few days but the generated code to get/set them is quite difficult to work with in the abstract. 

I also have not benchmarked this (yet?) - does that need to be done? I can do that before merge if necessary but I'm not clear on what the benchmarks are used for right now...


build.sbt
```diff
@@ -276,6 +277,24 @@ lazy val tensorflow: Project = project
     test % "test->test"
   )
 
+lazy val protobuf: Project = project
+  .in(file("protobuf"))
+  .settings(
+    commonSettings,
+    moduleName := "magnolify-protobuf",
+    description := "Magnolia add-on for Apache Avro",
```
nevillelyh:
```suggestion
    description := "Magnolia add-on for Google Protocol Buffer",
```

build.sbt
```diff
@@ -276,6 +277,24 @@ lazy val tensorflow: Project = project
...
+    libraryDependencies ++= Seq(
+      "com.github.os72" % "protoc-jar" % protobufVersion
```
nevillelyh: This jar pulls in the binary compiler which we don't need as part of the library.

You should be depending on `protobuf-java` instead as `Provided`?
https://github.com/nevillelyh/protobuf-generic/blob/master/build.sbt#L16

protobuf/src/main/protobuf/TestProto2.proto
```diff
@@ -0,0 +1,61 @@
+syntax = "proto2";
```
nevillelyh: These are for testing, should be in `src/test`. Not sure if the protobuf plugin supports that. Or you might have to split them into separate projects like in https://github.com/nevillelyh/protobuf-generic

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -0,0 +1,171 @@
...
+    override def from(v: Message): T = {
+      caseClass.construct { p =>
+        val fieldDescriptor = v.getDescriptorForType.findFieldByName(p.label)
+
+        if (fieldDescriptor.getType == FieldDescriptor.Type.MESSAGE) {
+          // nested
+          p.typeclass.fromAny(v.getField(fieldDescriptor))
+        } else {
+          // non-nested
+          p.typeclass.fromAny(v.getField(fieldDescriptor))
```
nevillelyh: These 2 branches look identical?

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -0,0 +1,171 @@
...
+        }
+      }
+    }
+
+    override def to(v: T, bu: Message.Builder): Message = {
+
+      // clear builder from previous runs before using it to construct a new instance
+      caseClass.parameters
+        .foldLeft(bu.clear()) { (b, p) =>
```
nevillelyh: This feels fishy. Looks like all calls to this method are sharing the same builder instance created at https://github.com/spotify/magnolify/pull/51/files#diff-aa33fce540e8b204301cecff009d7a0aR47

You can probably fix this by making `bu` a `() => Message.Builder`.

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -0,0 +1,171 @@
...
+    at[T, T, T](identity)(identity)
+
+  implicit val pfBoolean = id[Boolean]
+  implicit val pfString = id[String]
+  implicit val pfInt = id[Int]
+  implicit val pfLong = id[Long]
+  implicit val pfFloat = id[Float]
+  implicit val pfDouble = id[Double]
+  implicit val pfBytes =
+    from[Array[Byte], ByteBuffer](bb => ju.Arrays.copyOfRange(bb.array(), bb.position(), bb.limit())
```
nevillelyh: Why do we need to support `ByteBuffer`? It's quite error-prone. `ByteString` is a safer interface with Protobuf.

build.sbt
```diff
@@ -26,7 +26,7 @@ val datastoreVersion = "1.6.3"
 val guavaVersion = "28.2-jre"
 val jacksonVersion = "2.10.2"
 val jodaTimeVersion = "2.10.5"
-val protobufVersion = "3.10.0"
+val protobufVersion = "3.11.1" // needs to match protoc-jar version in project/plugins.sbt
```
nevillelyh: It shouldn't have to match if you depend on `protoc-jar` only for compiling `.proto` and not its transitive `protobuf-java` dependency.

build.sbt
nevillelyh: Also see https://github.com/nevillelyh/protobuf-generic/blob/master/project/plugins.sbt#L20

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
anne-decusatis: I copy/pasted this from somewhere, we don't have to support it. will update to ByteString :)

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
anne-decusatis: It was a deliberate choice to not have to do the (slow?) reflection within every step. I should probably do it both ways and benchmark, I will make the rest of the fixes and then look into that.

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
anne-decusatis: oops, fixing

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
nevillelyh: Was gonna ask about reflection & performance next ;)
Problem with mutable state like this is potential concurrency bugs, e.g. when the same `ProtobufType` is shared by multiple threads.

To reduce reflection overhead you can probably cache the reflected method from `implicitly[ClassTag[ParentMsgT]].runtimeClass.getMethod("newBuilder")`.

protobuf/src/main/protobuf/TestProto2.proto
anne-decusatis: plugin doesn't seem to support it, I made a downstream project protobufTest for them

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
anne-decusatis: it sounds like this is probably the 'wrong way' to do it because of concurrency regardless of performance, then, thanks! I will find a way to update that. I'm also still not entirely sure what's required in terms of benchmarking though, advice welcome.

anne-decusatis: it looks like I can't cache the reflected method at the object level because `Method` isn't serializable. I can create a new builder on this line from the existing one, instead of clearing the existing one. I'll push that, let me know what you think.

nevillelyh: There should be ways around serialization issue but we don't have to prematurely optimize without benchmarks, which you can plugin here: https://github.com/spotify/magnolify/blob/master/jmh/src/test/scala/magnolify/jmh/MagnolifyBench.scala

anne-decusatis:
I pasted some benchmarking results here: https://gist.github.com/anne-decusatis/18a8de7d839d083c7996c020f31c6bf7

In short, it looks like `from` has similar performance to Avro, but `to` is about 6x slower. Do you think more optimization is needed @nevillelyh ?
```
[info] Benchmark                Mode  Cnt      Score      Error  Units
[info] ProtobufBench.protoFrom  avgt   20  16212.446 ± 1027.635  ns/op
[info] ProtobufBench.protoTo    avgt   20  61193.899 ± 6813.971  ns/op
```
compared to
```
[info] Benchmark             Mode  Cnt      Score      Error  Units
[info] AvroBench.avroFrom    avgt   20  15695.917 ±  393.710  ns/op
[info] AvroBench.avroSchmea  avgt   20      6.847 ±    0.781  ns/op
[info] AvroBench.avroTo      avgt   20  10080.885 ± 2706.006  ns/op
```

anne-decusatis:
also looks like the travis build is failing with ```java.lang.RuntimeException: error occurred while compiling protobuf files: Cannot run program "protoc": error=2, No such file or directory``` - not sure how to reproduce that locally :(

README.md
```diff
@@ -131,6 +132,44 @@ val exampleBuilder: Example.Builder = exampleType.to(record)
 val copy = exampleType.from(exampleBuilder.build)
 ```
 
+Protobuf works slightly differently in that you need to specify both the type of the case class
```
nevillelyh: Every `*Type` works differently, some a lot more fundamentally different than others e.g. `ExampleField` that's not explainable in 2 paragraphs without prior hands on experience.
Also documenting (un)supported features has a very high chance of going out of sync with code.

nevillelyh:
Messed up the rebase :(
Gonna close this one and squash/merge the other instead.

PR #65, opened 2020-02-25
add ProtobufType

nevillelyh:


build.sbt
```diff
@@ -23,11 +23,10 @@ val avroVersion = "1.9.2"
 val bigqueryVersion = "v2-rev20191211-1.30.3"
 val catsVersion = "2.0.0"
 val datastoreVersion = "1.6.3"
-val grpcApiVersion = "1.17.0"
```
nevillelyh: Why was this needed? Everything works fine for me without it.

build.sbt
```diff
@@ -23,11 +23,10 @@ val avroVersion = "1.9.2"
...
 val guavaVersion = "28.2-jre"
 val jacksonVersion = "2.10.2"
 val jodaTimeVersion = "2.10.5"
-val protobufVersion = Option(sys.props("protobuf.version")).getOrElse("3.11.3")
```
nevillelyh: Keeping this simple for now since we're not trying to cover a cube of protoc/proto2/proto3 versions. We can keep this latest and `protoc-jar` to whatever available since it's usually a few versions behind.

build.sbt
```diff
@@ -149,8 +147,12 @@ lazy val test: Project = project
     libraryDependencies ++= Seq(
       "org.scalacheck" %% "scalacheck" % scalacheckVersion % Test,
       "org.typelevel" %% "cats-core" % catsVersion % Test
+    ),
+    protobufRunProtoc in ProtobufConfig := (args =>
```
nevillelyh: Moving the test `.proto`s to `test` to keep the project structure simple.

build.sbt
```diff
@@ -332,16 +315,8 @@ lazy val jmh: Project = project
       "com.google.apis" % "google-api-services-bigquery" % bigqueryVersion % Test,
       "joda-time" % "joda-time" % jodaTimeVersion % Test,
       "com.google.cloud.datastore" % "datastore-v1-proto-client" % datastoreVersion % Test,
-      "com.google.api.grpc" % "proto-google-common-protos" % grpcApiVersion % Test,
       "org.tensorflow" % "proto" % tensorflowVersion % Test
-    ),
-    // proto config
-    protobufProtocOptions in ProtobufConfig ++= Seq("--include_std_types"),
```
nevillelyh: What are these flags for? Works fine without them.

project/plugins.sbt
```diff
@@ -1,7 +1,11 @@
...
 addSbtPlugin("com.github.gseitz" % "sbt-release" % "1.0.13")
 addSbtPlugin("com.jsuereth" % "sbt-pgp" % "2.0.1")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.3.1")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.6.1")
 addSbtPlugin("org.xerial.sbt" % "sbt-sonatype" % "3.8.1")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.3.7")
-addSbtPlugin("com.github.gseitz" % "sbt-protobuf" % "0.6.5")
+
+libraryDependencies ++= Seq(
+  "com.github.os72" % "protoc-jar" % "3.11.1"
```
nevillelyh: We only need the `protoc` compiler binary from this jar. It's usually a few versions behind but fine.

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -29,26 +29,26 @@ import scala.annotation.implicitNotFound
 import scala.language.experimental.macros
 import scala.reflect.ClassTag
 
-sealed trait ProtobufType[T, ParentMsgT <: Message] extends Converter[T, ParentMsgT, ParentMsgT] {
```
nevillelyh: For some reason "`ParentMsgT` extends `Message`" just feels counter-intuitive.

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -117,32 +111,39 @@ object ProtobufField {
 
   def apply[T](implicit f: ProtobufField[T]): ProtobufField[T] = f
 
-  //////////////////////////////////////////////////
+  def from[T]: FromWord[T] = new FromWord[T]
```
nevillelyh: Same trick as in `AvroType`, etc. breaking away so only `T` needs to be supplied and `U` can be inferred from `f`.

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -29,26 +29,26 @@ import scala.annotation.implicitNotFound
...
 
 object ProtobufType {
-  implicit def apply[T, ParentMsgT <: Message: ClassTag](
-    implicit f: ProtobufField.Record[T]
-  ): ProtobufType[T, ParentMsgT] =
-    new ProtobufType[T, ParentMsgT] {
-      override def from(v: ParentMsgT): T = f.from(v)
-      override def to(v: T): ParentMsgT =
+  implicit def apply[T, MsgT <: Message](
+    implicit f: ProtobufField.Record[T], ct: ClassTag[MsgT]
```
nevillelyh: I tend to add `implicit` argument if it needs to be accessed via `implicitly` in the method anyway.

protobuf/src/test/scala/magnolify/protobuf/test/ProtobufTypeSpec.scala
```diff
@@ -26,8 +26,8 @@ import magnolify.cats.auto._
 import magnolify.scalacheck.auto._
 import magnolify.protobuf._
 import magnolify.shims.JavaConverters._
-import magnolify.skeleton.proto.TestProto2._
-import magnolify.skeleton.proto.TestProtoTypes._
+import magnolify.test.Proto2._
+import magnolify.test.Proto3._
```
nevillelyh: Looks like the main difference is 2 vs 3, so these names make more sense?

protobuf/src/test/scala/magnolify/protobuf/test/ProtobufTypeSpec.scala
```diff
@@ -37,20 +37,16 @@ import scala.reflect._
 object ProtobufTypeSpec extends MagnolifySpec("ProtobufRecordType") {
   private def test[T: ClassTag: Arbitrary: Eq, U <: Message: ClassTag](
     implicit tpe: ProtobufType[T, U],
-    eqMessage: Eq[U] = Eq.instance[U] { (first, second) =>
-      first.getDescriptorForType == second.getDescriptorForType &&
-      (first.getAllFields.asScala == second.getAllFields.asScala)
-    }
+    eqt: Eq[T],
+    eqm: Eq[U] = Eq.instance[U](_.toByteString == _.toByteString)
```
nevillelyh: Since we can't chain implicits for leave fields, might as well compare serialized bytes, which is pretty deterministic for protos.

test/src/main/protobuf/Proto3.proto
```diff
@@ -1,25 +1,25 @@
 syntax = "proto3";
 
-package magnolify.skeleton.proto;
+package magnolify.test;
 
-option java_package = "magnolify.skeleton.proto";
+option java_package = "magnolify.test";
 
 message IntegersP3 {
-     int32 i = 1;
```
nevillelyh: Indentations

build.sbt
anne-decusatis: I was seeing issues where a class in the package `com.google.type` was causing a NoClassDefFoundError in the datastore package - would fail when I tried to run all tests locally. Adding this jar which contains that class fixed it for me. see https://github.com/spotify/magnolify/pull/51/commits/e098934305400a5c4e950864137ffedc6736aa50

build.sbt
anne-decusatis: fine with me, I couldn't get it to work locally that way :/

build.sbt
anne-decusatis: saw it in another example, fine with me to remove

project/plugins.sbt
anne-decusatis: what does this do? I read https://www.scala-sbt.org/1.x/docs/Plugins.html and I think, from that, that this allows us to use protoc-jar stuff in build.sbt - which doesn't seem necessary, since we compile the protos in each module and not in the build? I think I am missing something

build.sbt
nevillelyh: Yeah weird. I saw the dependency, but never had any problem with datastore without it.

build.sbt
nevillelyh: Probably because of the `protobufVersion` variable which points to a version not available in `protoc-jar` yet.

project/plugins.sbt
nevillelyh: Protobuf compiler, i.e. `.proto` -> `.java`, is native binary (C++) and not part of `sbt-protobuf`. `sbt-protobuf` calls whatever `protoc` in `PATH` by default.
`protoc-jar` packages pre-built binaries for multiple platforms.

nevillelyh:
Added `Option` support for `PROTO2` and a safety check. The check is runtime but at `implicit` callsite so not too terrible. It's possibly doable as compile time with a macro but IMO overkill.

nevillelyh:
Everything works now. Benchmark is very slow though, 24000 ns/op both from/to vs ~666/2400 for Avro.

protobuf/src/test/scala/magnolify/protobuf/test/ProtobufTypeSpec.scala
```diff
@@ -0,0 +1,106 @@
...
+
+import scala.reflect._
+
+object ProtobufTypeSpec extends MagnolifySpec("ProtobufRecordType") {
+  private def test[T: ClassTag: Arbitrary: Eq, U <: Message: ClassTag](
+    implicit tpe: ProtobufType[T, U],
+    eqt: Eq[T],
+    eqm: Eq[U] = Eq.instance[U](_.toByteString == _.toByteString)
+  ): Unit = {
+    ensureSerializable(tpe)
```
ClaireMcGinty: Were you able to verify (just ad hoc) that Beam's `ProtoCoder` is able to encode/decode these? Or even just test that the generated Message classes can roundtrip using`Message#WriteTo` and `Message#parseFrom (since that's what ProtoCoder does)?

nevillelyh: Yeah it works. Proto ser/de is much simpler, if you can build it it's guaranteed to build.
Also `eqm` is copied from `AvroType` to deal with bad `GenericRecord` clone which should be impossible here. I'll remove.

But maybe we can add a ser/de check of avro/proto/example records for every `*Type` later.

nevillelyh:
Will merge this and do micro-optimizations in a follow up PR.

PR #66, opened 2020-02-27
micro-optimize ProtobufType

nevillelyh:
```
# before
[info] Benchmark                Mode  Cnt     Score   Error  Units
[info] ProtobufBench.protoFrom  avgt       6671.790          ns/op
[info] ProtobufBench.protoTo    avgt       8377.487          ns/op

# after
[info] Benchmark                Mode  Cnt     Score   Error  Units
[info] ProtobufBench.protoFrom  avgt       2117.371          ns/op
[info] ProtobufBench.protoTo    avgt       3311.638          ns/op
```

### DIFF
--- a/protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
+++ b/protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
@@ -16,9 +16,10 @@
  */
 package magnolify.protobuf
 
+import java.lang.reflect.Method
 import java.{util => ju}
 
-import com.google.protobuf.Descriptors.FileDescriptor.Syntax.PROTO2
+import com.google.protobuf.Descriptors.FileDescriptor.Syntax
 import com.google.protobuf.Descriptors.{Descriptor, FieldDescriptor}
 import com.google.protobuf.{ByteString, Message}
 import magnolia._
@@ -27,6 +28,7 @@ import magnolify.shims.FactoryCompat
 import magnolify.shims.JavaConverters._
 
 import scala.annotation.implicitNotFound
+import scala.collection.concurrent
 import scala.language.experimental.macros
 import scala.reflect.ClassTag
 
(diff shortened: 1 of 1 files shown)


protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -107,21 +122,23 @@ object ProtobufField {
...
+      caseClass.parameters
+        .foldLeft(bu) { (b, p) =>
+          val field = fields(p.index)
           val value = if (field.getType == FieldDescriptor.Type.MESSAGE) {
             // nested records
             p.typeclass.to(p.dereference(v), b.newBuilderForField(field))
           } else {
             // non-nested
-            p.typeclass.to(p.dereference(v), b)
+            p.typeclass.to(p.dereference(v), null)
```
nevillelyh: Unrelated to the optimization, but builder is unused for primitives, setting it `null` to fail early.

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -143,7 +160,7 @@ object ProtobufField {
         override type ToT = pf.ToT
         override val hasOptional: Boolean = pf.hasOptional
         override def from(v: FromT): U = f(pf.from(v))
-        override def to(v: U, b: Message.Builder): ToT = pf.to(g(v), b)
+        override def to(v: U, b: Message.Builder): ToT = pf.to(g(v), null)
```
nevillelyh: Same as https://github.com/spotify/magnolify/pull/66/files#diff-aa33fce540e8b204301cecff009d7a0aL120

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -89,16 +91,29 @@ object ProtobufField {
...
+
     override val hasOptional: Boolean = caseClass.parameters.exists(_.typeclass.hasOptional)
 
     override def from(v: Message): T = {
       val descriptor = v.getDescriptorForType
       val syntax = descriptor.getFile.getSyntax
+      val fields = getFields(descriptor)
 
       caseClass.construct { p =>
-        val field = descriptor.findFieldByName(p.label)
```
nevillelyh: Looks like `findFieldByName` is O(n) internally so this would make a bigger difference for fat objects.

Issue #81, opened 2020-03-18
Add BigtableType for Mutation/Row


nevillelyh:
Where `Mutation` is the writer & `Row` reader type. Should be similar to `ExampleType` since neither support nested structure.

PR #82, opened 2020-03-19
Add BigtableType for Mutation/Row #81

stormy-ua:
Added support for Bigtable type derivation:

- converts an entity into a collection of Mutations
- reads an entity back from a BigTable row


nevillelyh:
Looks good! Thanks!

PR #84, opened 2020-03-21
BigtableType cleaned up. Additional writer/reader settings moved out to builders.

stormy-ua:
Bigtable interface cleaned up and now conforms to ```Converter[T, Reader, Writer]``` interface.

In addition to ```to/from``` functions coming from ```Converter```, ```BigtableType``` exposes writer and reader builders, which could be used to append additional settings to writer/reader;

```scala
  val mutations = BigtableType[Required]
    .toBuilder
    .withColumnFamily("cf")
    .withTimestamp(123L)
    .build
    .to(Required(true, 2, "a"))

  val row = mutationsToRow(mutations)

  val copy = BigtableType[Required]
    .fromBuilder
    .withColumnFamily("cf")
    .build
    .from(row)
```

If there is no need for an additional configuration then ```BigtableType``` could used directly e.g.:

```scala
  val mutations = BigtableType[Required]
    .to(Required(true, 2, "a"))

  val row = mutationsToRow(mutations)

  val copy = BigtableType[Required]
    .from(row)
```

### DIFF
--- a/bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
+++ b/bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
@@ -26,18 +26,62 @@ import scala.collection._
 import com.google.bigtable.v2.Mutation
 import com.google.bigtable.v2.Mutation.SetCell
 import com.google.cloud.bigtable.data.v2.models.{Row, RowCell}
+import magnolify.shared.Converter
 
 import scala.annotation.implicitNotFound
 import scala.language.experimental.macros
 import magnolify.shims.JavaConverters._
 
-sealed trait BigtableType[T] extends Serializable {
+sealed trait BigtableTypeReader[T] extends Serializable {
+  def from(v: Row): T
+}
 
-  def from(v: Row, columnFamily: String = BigtableType.DEFAULT_COLUMN_FAMILY_NAME): T
+sealed trait BigtableTypeWriter[T] extends Serializable {
+  def to(v: T): Iterable[Mutation]
+}
 
-  def to(v: T, columnFamily: String = BigtableType.DEFAULT_COLUMN_FAMILY_NAME,
-               timestamp: Long = 0L,
-               columnQualifier: String = null): Iterable[Mutation]
+trait BigtableType[T] extends Converter[T, Row, Iterable[Mutation]] {
+
+  def from(v: Row): T = fromBuilder.build.from(v)
+  def to(v: T): Iterable[Mutation] = toBuilder.build.to(v)
+
+  val toBuilder: BigtableTypeWriterBuilder[T]
+  val fromBuilder: BigtableTypeReaderBuilder[T]
+
+}
+
+private[bigtable] final class BigtableTypeReaderBuilder[T](f: BigtableField[T],
+                                                           columnFamily: String)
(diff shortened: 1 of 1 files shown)


nevillelyh:
Oops didn't realize you were already working on this. I cleaned up a bit in #85. Will comment there.

stormy-ua:
> Oops didn't realize you were already working on this. I cleaned up a bit in #85. Will comment there.

I was trying to focus only on these enhancements:

*  decouple additional parameters (columnFamily, timestamp etc.) from Converter functions (to/from) 
* expose a familiar builder-like interface through ```BigtableType```, so a user can optionally append an additional set of parameters if needed
* split settings for reader and writer e.g. timestamp isn't being used now for reading (```from```) so it would be confusing for a user to provide it without seeing any effect. In addition, I feel like there might requests for additional reader/writer parameters to add and they might not be the same for both reader and writer.

nevillelyh:
What about making the writer type `Seq[Mutation.Builder]` instead of `Seq[Mutation]` like other protobuf based converters like [`ExampleType`](https://github.com/spotify/magnolify/blob/master/tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala#L31)?

That way we don't have to expose any additional parameters, since the user can just chain additional builder methods and `.build()` when reader, e.g.: `xs.map(_.setTimestampMicros(.).build()`.

stormy-ua:
I like the idea of building `Mutation.Builder` instead of Mutation! It doesn’t solve the issue with the reader though, because we are reading from a Row and need to know familyName and other query params in advance. Maybe it might be a good idea to expose `Seq[Mutation.Builder]` directly for writer and some “query builder” for reader e.g. ```
BigtableType[T].query.withFamilyName(“f”).from(row)
```?

nevillelyh:
Yeah in my PR I lifted `columnFamily` & `timestampMicros` to `def apply` method assuming that all rows of the same type `T` should be stored under the same column family.

OTOH we can go even further, replacing `Row` with `List[RowCell]`, moving the responsibility of column family logic to user code. In addition, since we're talking about `Seq[Mutation]` with` Seq[Mutation.Builder]`, even the `timestampMicros` logic can be move out of the converter and left to user.

By doing this we can basically decouple any storage logic from `BigtableType`, and only offer core data type conversion logic, which is more consistent with the other converters, WDYT?

nevillelyh:
Updated #85 again. I'm now using `java.util.List[RowCell]` & `Seq[SetCell.Builder]` for reader/writer types to allow most control inside the conversion. Handling of column family & timestamp is now done at `BigtableType#apply` layer which also converts reader/writer types to `Row` & `Seq[Mutation]` for user consumption.

stormy-ua:
> Updated #85 again. I'm now using `java.util.List[RowCell]` & `Seq[SetCell.Builder]` for reader/writer types to allow most control inside the conversion. Handling of column family & timestamp is now done at `BigtableType#apply` layer which also converts reader/writer types to `Row` & `Seq[Mutation]` for user consumption. Any additional argument can be handled there as well.

Yeah, already saw your updates and left comments there. Will be closing this PR for now. Looks like yours already has the minimum we need to be enough flexible without exposing more complex/builder API.

PR #85, opened 2020-03-21
clean up BigtabeType

nevillelyh:


README.md
```diff
@@ -27,10 +27,10 @@ This library includes the following modules.
 
 - `magnolify-avro` - conversion between Scala types and [Apache Avro](https://github.com/apache/avro) `GenericRecord`
 - `magnolify-bigquery` - conversion between Scala types and [Google Cloud BigQuery](https://cloud.google.com/bigquery/) `TableRow`
+- `magnolify-bigtable` - conversion between Scala types and [Google Cloud Bigtable](https://cloud.google.com/bigtable) to `Mutation`, from `Row`
```
nevillelyh: Just to keep module sorted alphabetically.

README.md
```diff
@@ -114,6 +114,15 @@ val copy: Outer = tableRowType.from(tableRow)
 
 tableRowType.schema // BigQuery TableSchema
 
+// Bigtable Example
```
nevillelyh: Sorting sections by module name.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -14,88 +14,130 @@
...
-import magnolify.shims.JavaConverters._
-
-sealed trait BigtableType[T] extends Serializable {
-
-  def from(v: Row, columnFamily: String = BigtableType.DEFAULT_COLUMN_FAMILY_NAME): T
 
-  def to(v: T, columnFamily: String = BigtableType.DEFAULT_COLUMN_FAMILY_NAME,
-               timestamp: Long = 0L,
-               columnQualifier: String = null): Iterable[Mutation]
+sealed trait BigtableType[T] extends Converter[T, Row, Seq[Mutation]] {
```
nevillelyh: `BigtableType` can extend `Converter` now that I lifted `columnFamily` & `timestampMicros` as `val` instead of arguments.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -14,88 +14,130 @@
...
+        Mutation
+          .newBuilder()
+          .setSetCell(
+            b.setFamilyName(columnFamily).setTimestampMicros(timestampMicros)
+          )
+          .build()
+      }
   }
+
+  def mutationsToRow(key: ByteString, mutations: Seq[Mutation]): Row =
```
nevillelyh: Lifing `Seq[Mutation]` <=> `Row` conversion methods here since they're pretty handy? Also make mutations a `Seq` to avoid `Iterable` <=> `List` copies.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -14,88 +14,130 @@
...
 
 sealed trait BigtableField[T] extends Serializable {
   def get(columnFamily: String, v: Row, k: String): T
-  def put(k: String, v: T): Iterable[SetCell.Builder]
+  def put(k: String, v: T): Seq[SetCell.Builder]
 }
 
 object BigtableField {
   trait Primitive[T] extends BigtableField[T] {
-    type ValueT
```
nevillelyh: This `ValueT` is never used anywhere.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -109,112 +151,53 @@ object BigtableField {
...
-
-  implicit val btString = new Primitive[String] {
-    type ValueT = String
-
-    def fromByteString(v: ByteString): String =
-      v.toStringUtf8
-
-    def toByteString(v: String): ByteString =
-      ByteString.copyFromUtf8(v)
+  private def primitive[T](
```
nevillelyh: Added some more syntactic sugar to reduce boilerplate.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -109,112 +151,53 @@ object BigtableField {
...
+  implicit val btfByteString = new Primitive[ByteString] {
+    override def fromByteString(v: ByteString): ByteString = v
+    override def toByteString(v: ByteString): ByteString = v
   }
+  implicit val btfByteArray = from[ByteString](_.toByteArray)(ByteString.copyFrom(_))
+  implicit val btfString = from[ByteString](_.toStringUtf8)(ByteString.copyFromUtf8(_))
+
+  implicit def btfOption[A](implicit btf: BigtableField[A]): BigtableField[Option[A]] =
+    new BigtableField[Option[A]] {
+      override def get(columnFamily: String, v: Row, k: String): Option[A] = {
```
nevillelyh: Not sure how much difference it makes really, linear search for both `$k` and `$k.` vs binary search for one.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
stormy-ua: 👍

bigtable/src/test/scala/magnolify/bigtable/test/BigtableTypeSpec.scala
```diff
@@ -14,77 +14,64 @@
...
+      val mutations = tpe(t)
+      val row = BigtableType.mutationsToRow(ByteString.EMPTY, mutations)
+      val copy = tpe(row)
+      val rowCopy = BigtableType.mutationsToRow(ByteString.EMPTY, BigtableType.rowToMutations(row))
 
-      val copy = tpe.from(row)
-      eq.eqv(t, copy)
+      Prop.all(
+        eq.eqv(t, copy),
+        row == rowCopy
```
nevillelyh: Also test mutations <=> `Row` round trip.

bigtable/src/test/scala/magnolify/bigtable/test/BigtableTypeSpec.scala
```diff
@@ -14,77 +14,64 @@
...
+    implicit val arbByteString: Arbitrary[ByteString] =
+      Arbitrary(Gen.alphaNumStr.map(ByteString.copyFromUtf8))
+    implicit val eqByteString: Eq[ByteString] = Eq.instance(_ == _)
+    implicit val eqByteArray: Eq[Array[Byte]] = Eq.by(_.toList)
+    test[BigtableTypes]
+  }
 
-  implicit val efDuration: BigtableField.Primitive[Duration] =
-    BigtableField.from[Long](Duration.ofMillis)(_.toMillis)
+  {
```
nevillelyh: Move into a scope to avoid leaking implicits.

bigtable/src/test/scala/magnolify/bigtable/test/BigtableTypeSpec.scala
```diff
@@ -14,77 +14,64 @@
...
-
-  implicit val efUri: BigtableField.Primitive[URI] =
-    BigtableField.from[String](x => URI.create(x))(_.toString)
+  {
+    implicit val arbByteString: Arbitrary[ByteString] =
+      Arbitrary(Gen.alphaNumStr.map(ByteString.copyFromUtf8))
+    implicit val eqByteString: Eq[ByteString] = Eq.instance(_ == _)
+    implicit val eqByteArray: Eq[Array[Byte]] = Eq.by(_.toList)
+    test[BigtableTypes]
+  }
```
nevillelyh: Cover `ByteString` and `Array[Byte]`.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
stormy-ua: oh yeah, this is a leftover form ExampleType

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
stormy-ua: yeah, 2*log(n) should be better that n, but usually there are not so many fields so probably not critical.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -14,88 +14,133 @@
...
   val DEFAULT_COLUMN_FAMILY_NAME = "f"
+  val DEFAULT_COLUMN_QUALIFIER_NAME = "q"
 
-  implicit def apply[T](implicit f: BigtableField[T]): BigtableType[T] = new BigtableType[T] {
+  implicit def apply[T](implicit f: BigtableField[T]): BigtableType[T] =
+    BigtableType(BigtableType.DEFAULT_COLUMN_FAMILY_NAME, 0L)
 
-    def from(v: Row, columnFamily: String = BigtableType.DEFAULT_COLUMN_FAMILY_NAME): T =
-      f.get(columnFamily, v, null)
+  def apply[T](_columnFamily: String, _timestampMicros: Long)(
```
stormy-ua: sometimes only one parameter should be overridden. Maybe add default values, so only one could specified, but I think there might be some issues with the apply function - it might not compile.

stormy-ua:
LGTM! Left one small comment about adding default values to ```Bigtable.apply(columnFamily, timestamp)```, so only one could be specified at a time.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -14,88 +14,112 @@
...
-sealed trait BigtableType[T] extends Serializable {
-
-  def from(v: Row, columnFamily: String = BigtableType.DEFAULT_COLUMN_FAMILY_NAME): T
-
-  def to(v: T, columnFamily: String = BigtableType.DEFAULT_COLUMN_FAMILY_NAME,
-               timestamp: Long = 0L,
-               columnQualifier: String = null): Iterable[Mutation]
 
+sealed trait BigtableType[T] extends Converter[T, java.util.List[RowCell], Seq[SetCell.Builder]] {
+  def apply(v: Row, columnFamily: String): T = from(v.getCells(columnFamily))
```
nevillelyh: Are these good enough? We can reuse the same `BigtableType[T]` for `T` & different column families and timestamp settings. And if there aren't any other arguments we can live with a single method with default `timestampMicros`?

stormy-ua: I think this is good enough. Can't think of any example when it wouldn't work.

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -164,22 +162,82 @@ object BigtableField {
...
+
+      override def put(k: String, v: Option[A]): Seq[SetCell.Builder] =
+        v.toSeq.flatMap(btf.put(k, _))
+    }
+}
+
+private object RowCells {
+  private val cmp = ByteStringComparator.INSTANCE
+
+  private def find(
```
stormy-ua: Just wanted to mention that searching a cell in a list of RowCells wouldn't be that efficient as looking up a rowCell directly in a row using Row.getCells(), but looks like you already solved this issue :)

stormy-ua:
I feel like the updated implementation is what we need. It might put more constraints for BigtaleType query/reader API than needed, but this should be fine until we learn what other parameters could/should be used for querying.

PR #88, opened 2020-03-29
BigtableType reader Row type changed to the one being used by scio BigtableIO (needed for seamless integration with scio)

stormy-ua:
I was creating [scio examples for `magnolify` BigtableType](https://github.com/spotify/scio/pull/2816) when I realized that Bigtable row type we are using in `magnolify` `com.google.cloud.bigtable.data.v2.models.Row` doesn't match row type used by `scio`. `scio` uses `com.google.bigtable.v2.Row` in [scio syntax](https://github.com/spotify/scio/blob/master/scio-bigtable/src/main/scala/com/spotify/scio/bigtable/syntax/ScioContextSyntax.scala#L48) because under the hood Beam uses it in [BigtableIO](https://github.com/apache/beam/blob/master/sdks/java/io/google-cloud-platform/src/main/java/org/apache/beam/sdk/io/gcp/bigtable/BigtableIO.java#L211).

I have migrated BigtableType to `com.google.bigtable.v2.Row` for seamless integration of `magnolify` and `scio`.

### DIFF
--- a/bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
+++ b/bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
@@ -18,10 +18,9 @@ package magnolify.bigtable
 
 import java.nio.ByteBuffer
 
-import com.google.bigtable.v2.Mutation
+import com.google.bigtable.v2.{Cell, Column, Family, Mutation, Row}
 import com.google.bigtable.v2.Mutation.SetCell
 import com.google.cloud.bigtable.data.v2.internal.ByteStringComparator
-import com.google.cloud.bigtable.data.v2.models.{Row, RowCell}
 import com.google.protobuf.ByteString
 import magnolia._
 import magnolify.shared.Converter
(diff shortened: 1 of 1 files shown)


nevillelyh:
LGTM, I also removed `google-cloud-bigtable` dependency since it pulls in a lot of conflicting Guava/gRPC/etc. deps and has caused problem in the past. I had to fork the `ByteStringComparator` though :man_shrugging: 

Also changed the `xs.foldLeft(builder)(_.addColumn(_))` patterns to `builder.addAllColumns(xs)`, mostly for readability.

https://github.com/spotify/magnolify/commit/fec15cee0f1b46bb76efd93362b01da20d4e6530

stormy-ua:
@nevillelyh I was also looking into removing it, but didn't want to fork the comparator, but I think in  general I agree - it is better to reduce dependencies as much as possible to alleviate the pain with transitive dependencies. `builder.addAllColumns(xs)` - nice! how could I missed that :)

PR #114, opened 2020-04-29
add @asNone annotation for PROTO3

nevillelyh:


protobuf/src/test/scala/magnolify/protobuf/test/ProtobufTypeSpec.scala
```diff
@@ -109,3 +131,16 @@ case class NestedNoOption(
   r: Required,
   l: List[Required]
 )
+
+case class NullableNoneValue(
+  @noneValue(true) b: Option[Boolean],
```
nevillelyh, line 136: Is there a better name? `@none`/`@asNone`?

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -187,23 +225,45 @@ object ProtobufField {
...
+  private class OptionProtobufField[T: ClassTag](val f: ProtobufField[T])
+      extends ProtobufField[Option[T]] {
+    override type FromT = f.FromT
+    override type ToT = f.ToT
+    override def from(v: f.FromT): Option[T] = if (v == null) None else Some(f.from(v))
+    override def to(v: Option[T], b: Message.Builder): f.ToT = v match {
+      case None    => null.asInstanceOf[f.ToT]
+      case Some(x) => f.to(x, b)
+    }
+    def typeCheck(v: Any, name: String): Unit = {
```
anne-decusatis, line 237: please add doc comment here, separate from my request above to comment the code where this is called

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -187,23 +225,45 @@ object ProtobufField {
...
+      val vc = v.getClass
+      val cls = implicitly[ClassTag[T]].runtimeClass
+      val compatible = cls match {
+        case c if c == classOf[Boolean] => vc == classOf[java.lang.Boolean]
```
anne-decusatis, line 241: where does this list come from?

protobuf/src/test/scala/magnolify/protobuf/test/ProtobufTypeSpec.scala
anne-decusatis: I like `asNone`? maybe `valueAsNone` or `defaultAsNone`?

build.sbt
```diff
@@ -17,7 +17,7 @@
 name := "magnolify"
 description := "A collection of Magnolia add-on modules"
 
-val magnoliaVersion = "0.16.0"
+val magnoliaVersion = "0.14.5"
```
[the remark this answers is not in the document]
nevillelyh: It's so that some users can try it out with current scio which is still on 0.14.5. Will remove once we release.

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -103,22 +110,30 @@ object ProtobufField {
...
+            null
+          } else {
+            v.getField(field)
+          }
+          p.typeclass.fromAny(value)
         } else {
-          v.getField(field)
+          val value = p.typeclass.fromAny(v.getField(field))
+          if (nv.isDefined) {
+            if (value == nv.asInstanceOf[p.PType]) None else value
```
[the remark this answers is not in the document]
nevillelyh: Yeah that's the goal but will also make case classes not round trip properly, that's why we can't test this _easily_ with ScalaCheck.

```scala
case class A(@noneValue(10) i: Option[Int])
```

`A(Some(10))` gets converted into protobuf with `i = 10` but round trips back as `A(None)`

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
nevillelyh: It's to deal with `scala.Int` vs `java.lang.Integer` etc. having different erasure. Will add comment.

protobuf/src/test/scala/magnolify/protobuf/test/ProtobufTypeSpec.scala
nevillelyh: I'm also leaning towards `asNone`, that way `None` is also properly capitalized :)
`default` can be confused with scala arg default value.

nevillelyh:
@anne-decusatis addressed your comments, WDYT?

nevillelyh:
This turned out to be a bad idea and really confusing for users. Abandoning.

Issue #108, opened 2020-04-24
Support default values in converters


nevillelyh:
- [x] Avro
- [x] BigQuery
- [x] Bigtable
- [x] Datastore
- [x] Protobuf
- [x] TensorFlow

```scala
scala> import magnolify.bigquery
final package bigquery
scala> import magnolify.bigquery._
import magnolify.bigquery._

scala> case class A(i: Long, j: Long = 10)
class A

scala> val t = TableRowType[A]
val t: magnolify.bigquery.TableRowType[A] = magnolify.bigquery.TableRowType$$anon$1@4b88be62

scala> import com.google.api.services.bigquery.model.TableRow
import com.google.api.services.bigquery.model.TableRow

scala> val tr = new TableRow().set("i", 123)
val tr: com.google.api.services.bigquery.model.TableRow = GenericData{classInfo=[f], {i=123}}

scala> t(tr)
java.lang.NullPointerException
  at magnolify.bigquery.TableRowField$.$anonfun$trfLong$1(TableRowType.scala:132)
  at magnolify.bigquery.TableRowField$.$anonfun$trfLong$1$adapted(TableRowType.scala:132)
  at magnolify.bigquery.TableRowField$$anon$4.from(TableRowType.scala:127)
  at magnolify.bigquery.TableRowField.fromAny(TableRowType.scala:61)
  at magnolify.bigquery.TableRowField.fromAny$(TableRowType.scala:61)
  at magnolify.bigquery.TableRowField$$anon$4.fromAny(TableRowType.scala:124)
  at magnolify.bigquery.TableRowField$$anon$2.$anonfun$from$1(TableRowType.scala:87)
  at $anon$1.construct(<console>:1)
  at $anon$1.construct(<console>:1)
  at magnolify.bigquery.TableRowField$$anon$2.from(TableRowType.scala:87)
  at magnolify.bigquery.TableRowField$$anon$2.from(TableRowType.scala:77)
  at magnolify.bigquery.TableRowType$$anon$1.from(TableRowType.scala:47)
  at magnolify.bigquery.TableRowType$$anon$1.from(TableRowType.scala:43)
  at magnolify.bigquery.TableRowType.apply(TableRowType.scala:37)
  at magnolify.bigquery.TableRowType.apply$(TableRowType.scala:37)
  at magnolify.bigquery.TableRowType$$anon$1.apply(TableRowType.scala:43)
  ... 40 elided
```

nevillelyh:
[...]

PR #115, opened 2020-05-02
Support default values for `{Bigtable,Entity,TableRow}Type`s #108

nevillelyh:


bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -127,8 +148,24 @@ object BigtableField {
     private def key(prefix: String, label: String): String =
       if (prefix == null) label else s"$prefix.$label"
 
-    override def get(xs: java.util.List[Column], k: String): T =
-      caseClass.construct(p => p.typeclass.get(xs, key(k, p.label)))
+    override def get(xs: java.util.List[Column], k: String): Result[T] = {
+      var fallback = true
```
stormy-ua, line 152: I m wondering if this could be implemented without `var`, but probably this isn't going to be a simple change. I checked `magnoliy` and not seeing any stateful `construct` equivalents.

nevillelyh: `caseClass.rawConstruct(fieldValues: Seq[Any])` so we can technically do a `caseClass.parameters.foldLeft()` with `z: (Seq[Any], Boolean)` but that feels less readable :)

Plus this state is never leaked and shouldn't cause threading issue.

stormy-ua: Yeah, I had an idea to fold over parameter, but didn't know there is a `rawConstruct` function. good to know! I agree, it would be over-complication w/o any significant benefits.

PR #126, opened 2020-05-14
Use ReadOnlyCaseClass for deriving readonly typeclasses

### LINKED ISSUE #125: Use ReadOnlyCaseClass for typeclasses that does not need to construct instances
See https://twitter.com/propensive/status/1247578833680773122

Will help reduce compile time as well as allowing those derivation to work with classes where the constructor is inaccessible :)

Migration involves changing `CaseClass` to `ReadOnlyCaseClass`

jatcwang:
Fixes #125


nevillelyh:
Fixed up in 852565900549108bd9087c3f20bd81e0b9764937. Thanks!

PR #152, opened 2020-06-29
Neville/case mapping

nevillelyh:


avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -43,12 +43,11 @@ sealed trait AvroType[T] extends Converter[T, GenericRecord, GenericRecord] {
 }
 
 object AvroType {
-  implicit def apply[T: AvroField.Record]: AvroType[T] = apply(identity: CaseMapper)
+  implicit def apply[T: AvroField.Record]: AvroType[T] = AvroType(identity: CaseMapper)
 
-  def apply[T: AvroField.Record](g: CaseMapper): AvroType[T] =
+  def apply[T](g: CaseMapper)(implicit f: AvroField.Record[T]): AvroType[T] =
```
nevillelyh, line 48: Explicit `(implicit f)` so that we don't have to do `val f = implicitly[T]` inside.

avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -89,9 +88,9 @@ object AvroField {
         false,
         caseClass.parameters.map { p =>
           new Schema.Field(
-            cm map p.label,
+            cm.map(p.label),
```
nevillelyh, line 91: I avoid infix for most cases, just for the sake of consistency. The only exceptions maybe `zip/union/++` where both sides are equivalent.

avro/src/test/scala/magnolify/avro/test/AvroTypeSpec.scala
```diff
@@ -149,6 +134,22 @@ object AvroTypeSpec extends MagnolifySpec("AvroType") {
     expectException[IllegalArgumentException](AvroType[DoubleFieldDoc]).getMessage ==
       "requirement failed: More than one @doc annotation: magnolify.avro.test.DoubleFieldDoc#i"
   )
+
+  {
```
nevillelyh, line 138: Not pretty but passable without pulling in ScalaTest which I've been punting on forever :shrug:
Also more explicitly checking the field names.

PR #197, opened 2020-09-29
change snake case examples to use underscores

freyrsae:
currently the snake case examples use hyphens which is non-custom for snake case.


nevillelyh:
Good catch thanks!

PR #215, opened 2020-10-20
add simple Refined implicits

nevillelyh:


refined/src/main/scala/magnolify/refined/package.scala
```diff
@@ -0,0 +1,65 @@
...
+
+import scala.language.implicitConversions
+
+package object refined {
+  object avro {
+    implicit def refinedAvro[T: AvroField, P](implicit v: Validate[T, P]): AvroField[Refined[T, P]] =
+      AvroField.from[T](unsafeRefine[T, P])(_.value)
+  }
+
+  object bigquery {
```
syodage, line 36: is this safe if `magnolify-bigquery` is not in the classpath?

nevillelyh: Yeah all dependencies are `Provided` so people will have to explicitly include `magnolify-bigquery` etc. This is so that we don't create a `magnolify-refined-*` for each one of Avro, BQ, Bigtable, etc.

syodage (approved):

PR #257, opened 2021-01-19
Update TensorFlow to new dependency that targets 2.3.1

regadas:


nevillelyh:
We need to bump version for this right? Since it's not compat with 1.x types.

regadas:
@nevillelyh right, that makes sense!

PR #262, opened 2021-02-02
Support for java.lang.UUID for Bigtable type

stormy-ua:
We have a use case when we need to read/write a case class with `java.lang.UUID` field from/to a BT table. The PR adds support for this type. It is better to store UUID in bytes representation, because it consumes less space than in its string form.

### DIFF
--- a/bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
+++ b/bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
@@ -190,11 +190,9 @@ object BigtableField {
   implicit val btfFloat = primitive[Float](java.lang.Float.BYTES)(_.getFloat)(_.putFloat(_))
   implicit val btfDouble = primitive[Double](java.lang.Double.BYTES)(_.getDouble)(_.putDouble(_))
   implicit val btfBoolean = from[Byte](_ == 1)(if (_) 1 else 0)
-  implicit val btfUUID = primitive[UUID](16)(bb =>
-    new UUID(bb.getLong, bb.getLong)){
-    (bb, uuid) =>
-      bb.putLong(uuid.getMostSignificantBits)
-      bb.putLong(uuid.getLeastSignificantBits)
+  implicit val btfUUID = primitive[UUID](16)(bb => new UUID(bb.getLong, bb.getLong)) { (bb, uuid) =>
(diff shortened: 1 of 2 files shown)


bigtable/src/test/scala/magnolify/bigtable/test/BigtableTypeSuite.scala
```diff
@@ -30,6 +29,7 @@ import magnolify.test.Simple._
 import magnolify.test._
 import org.scalacheck._
 
+import java.util.UUID
```
nevillelyh, line 32:
```suggestion
```

bigtable/src/test/scala/magnolify/bigtable/test/BigtableTypeSuite.scala
```diff
@@ -18,7 +18,6 @@ package magnolify.bigtable.test
 
 import java.net.URI
 import java.time.Duration
```
nevillelyh, line 20:
```suggestion
import java.time.Duration
import java.util.UUID

```

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -17,14 +17,14 @@
 package magnolify.bigtable
 
 import java.nio.ByteBuffer
```
nevillelyh, line 19:
```suggestion
import java.nio.ByteBuffer
import java.util.UUID

```

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -17,14 +17,14 @@
...
-
 import com.google.bigtable.v2.{Cell, Column, Family, Mutation, Row}
 import com.google.bigtable.v2.Mutation.SetCell
 import com.google.protobuf.ByteString
 import magnolia._
 import magnolify.shared._
 import magnolify.shims.JavaConverters._
 
+import java.util.UUID
```
nevillelyh, line 27:
```suggestion
```

nevillelyh (approved):
LGTM

It seems that IntelliJ changed default import formatter? Fixed here for consistency but wish there's a way to automate this. Scalafmt doesn't support imports.

Issue #255, opened 2021-01-05
Ideas


nevillelyh:
Just a brain dump of nice to haves. Nothing about feasibility though.

- Avro union <=> Either/sealed trait
- Protobuf oneof <=> Either/sealed trait
- Code-gen of case classes from Avro/BigQuery/Protobuf/Parquet schemas, optionally with field projection
- Proper fix for cats built-in implicit conflict e.g. `Option[T]`, `List[T]` that are also ADTs.

PR #284, opened 2021-02-24
support repeated leafs in BigtableType #255

nevillelyh:


bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -219,14 +224,66 @@ object BigtableField {
...
+    new BigtableField[Option[T]] {
+      override def get(xs: java.util.List[Column], k: String)(cm: CaseMapper): Value[Option[T]] =
         Columns.findNullable(xs, k).map(btf.get(_, k)(cm).toOption).getOrElse(Value.Default(None))
 
-      override def put(k: String, v: Option[A])(cm: CaseMapper): Seq[SetCell.Builder] =
+      override def put(k: String, v: Option[T])(cm: CaseMapper): Seq[SetCell.Builder] =
         v.toSeq.flatMap(btf.put(k, _)(cm))
     }
+
+  implicit def btfIterable[T, C[T]](implicit
```
stormy-ua, line 236: Nice! Now we can store iterables in a cell 🎉

bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -219,14 +224,66 @@ object BigtableField {
...
+          case Some(size) =>
+            val ba = new Array[Byte](size)
+            (1 to n).foreach { _ =>
+              buf.get(ba)
+              b += btf.fromByteString(ByteString.copyFrom(ba))
+            }
+          case None =>
+            (1 to n).foreach { _ =>
+              val s = buf.getInt
+              val ba = new Array[Byte](s)
```
stormy-ua, line 258: I wonder if we need to handle the case with an empty string somehow explicitly, but you have added a test for `Repeated` so looks like it should work.

nevillelyh: Yeah, arrays can be empty, and scalacheck should generate "" cases.

Issue #287, opened 2021-02-26
BigtableType field names shortening


stormy-ua:
Bigtable column qualifiers are [stored as data](https://cloud.google.com/bigtable/docs/schema-design#columns) for each cell, so the best option is to keep column names as short as possible. At the same time, in a user case class field names are supposed to be long enough to convey field designation. It would be great to annotate case class fields with an attribute having a shortened field name e.g.:

```scala
case class MyClass(@btfieldname("f") superDescriptiveField: String)
```

I was thinking about adding implicit ClassTag to the BigtableType apply function and use reflection to traverse the type and build fields map once and cache it. Wondering if there are any other thoughts how to do this.

nevillelyh:
This you can do already by abusing `CaseMapper`?
https://github.com/spotify/magnolify/blob/master/docs/bigtable.md

stormy-ua:
Yes :) I was thinking about renaming `CaseMapper` to `FieldNameMapper`. Oh nice, I didn't know magnolia supports annotations. Thanks for the hint!

nevillelyh:
You're right, `CaseMapper` is misleading since you can do whatever to the string 😭

nevillelyh:
Closing as won't fix.

PR #293, opened 2021-03-02
#287 BigtableType field names shortening

stormy-ua:
The PR adds support for the optional field name mapping/shortening with the `@fieldName` static annotation. More details about why this is needed could be found in [this issue](https://github.com/spotify/magnolify/issues/287).

### DIFF
--- a/bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
+++ b/bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
@@ -18,7 +18,6 @@ package magnolify.bigtable
 
 import java.nio.ByteBuffer
 import java.util.UUID
-
 import com.google.bigtable.v2.{Cell, Column, Family, Mutation, Row}
 import com.google.bigtable.v2.Mutation.SetCell
 import com.google.protobuf.ByteString
@@ -27,8 +26,12 @@ import magnolify.shared._
 import magnolify.shims._
 import magnolify.shims.JavaConverters._
 
-import scala.annotation.implicitNotFound
+import scala.annotation.{implicitNotFound, StaticAnnotation}
 import scala.language.experimental.macros
+
+class fieldName(fieldName: String) extends StaticAnnotation with Serializable {
+  override def toString: String = fieldName
+}
 
 sealed trait BigtableType[T] extends Converter[T, java.util.List[Column], Seq[SetCell.Builder]] {
   def apply(v: Row, columnFamily: String): T =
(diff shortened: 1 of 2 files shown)


nevillelyh:
Can we make it work with `CaseMapper` though? A little confusing with 2 overlapping ways of doing things.

This should work? It'd apply to all nested cases of `very_long_name` but shouldn't be a huge deal?
```scala
CaseMapper {
  case "very_long_name" => "short"
  case n => n
}
```

stormy-ua:
> Can we make it work with `CaseMapper` though? A little confusing with 2 overlapping ways of doing things.
> 
> This should work? It'd apply to all nested cases of `very_long_name` but shouldn't be a huge deal?
> 
> ```scala
> CaseMapper {
>   case "very_long_name" => "short"
>   case n => n
> }
> ```



> Can we make it work with `CaseMapper` though? A little confusing with 2 overlapping ways of doing things.
> 
> This should work? It'd apply to all nested cases of `very_long_name` but shouldn't be a huge deal?
> 
> ```scala
> CaseMapper {
>   case "very_long_name" => "short"
>   case n => n
> }
> ```

I was thinking about the approach with `CaseMapper`, but the problem is that we would need to hardcode the field name as a string and it might get out of sync if someone renames the field but forgets to update the hardcoded string with the name. Right, looks like two almost overlapping ways of doing this, but I don't think we can drop `CaseMapper` in favor of the `@fieldName`, because it wouldn't be possible to provide a generic case mapper then. At the same time, `@fieldName` is better for a task of simple field mapping, because we don't duplicate a field name in a hardcoded string. So I would say they overlapping, but in some sense complementing each other, but I leave it to you to make the final call :) We can definitely do the CaseMapper approach. It would be just less elegant.

nevillelyh:
The other possible source of confusion is which one wins, e.g. `@Fieldname("mF") myFieldName`, could end up being `mF` (annotation wins) or `m_f` (annotation + to_snake_case mapper).

Back to the root cause. What's stopping us from just using short names and comments/noop annotations as doc? Like:
```scala
case class Record(@doc("very long description") f: Int)
```

That's at least more inline with Avro even though `@doc` isn't persistent in schema or on disk.

stormy-ua:
> The other possible source of confusion is which one wins, e.g. `@Fieldname("mF") myFieldName`, could end up being `mF` (annotation wins) or `m_f` (annotation + to_snake_case mapper).
> 
> Back to the root cause. What's stopping us from just using short names and comments/noop annotations as doc? Like:
> 
> ```scala
> case class Record(@doc("very long description") f: Int)
> ```
> 
> That's at least more inline with Avro even though `@doc` isn't persistent in schema or on disk.

I think that's a good idea. The only problem is with IntelliSense or with the code readability accessing the field - it would be just `f`, but I agree the order they would be applied is confusing. Not sure what we can do here to avoid the confusion. Let's probably hold off with the PR - I will try to think how to make it better and less confusing.

What do you think about creating a case mapper using magnolia which can extract field names automatically? Then it wouldn't be that confusing because a user would need to explicitly choose to use auto-field name mapping caseMapper. The caseMapper itself could be built using magnolia I think.

stormy-ua:
Closing this for now. After the conversation with Neville it doesn't seem like adding more field/case mapping logic to the library is a good idea. Better to keep it separate from the lib.

PR #409, opened 2021-08-24
check Avro compatibility when reading Parquet

nevillelyh:


parquet/src/main/scala/magnolify/parquet/ParquetType.scala
```diff
@@ -102,13 +109,37 @@ object ParquetType {
...
         parquetType = SerializationUtils.fromBase64(context.getConfiguration.get(ReadTypeKey))
       }
+
+      val metadata = context.getKeyValueMetadata
+      val model = metadata.get(ParquetWriter.OBJECT_MODEL_NAME_PROP)
+      val isAvroFile = (model != null && model.contains("avro")) ||
+        metadata.containsKey(AVRO_SCHEMA_METADATA_KEY) ||
+        metadata.containsKey(OLD_AVRO_SCHEMA_METADATA_KEY)
+      if (isAvroFile) {
+        require(
```
nevillelyh, line 130: These `require`s might be too strict since the read/write should still work as long as we don't touch any repeated fields. Should they be warnings instead?

PR #454, opened 2021-11-12
Add support for required->optional backward compatibility

stormy-ua:
The PR adds support for backward compatible change of a field repetition from `required` to `optional`. The case is supported by Avro according to [docs](https://avro.apache.org/docs/current/spec.html): 

```
if reader's is a union, but writer's is not:
The first schema in the reader's union that matches the writer's schema is recursively resolved against it. If none match, an error is signalled.
```

It is also supported by Parquet itself bc `optional` always able to parse `required` bc it is always there. I added a bunch of tests + tested in my own sandbox and didn't find any issues with that.

### Why is this needed ?!

This is needed to support a reader schema that has a field with repetition changed from `required` to `optional`. The change itself is fully supported by Avro and the Parquet itself if backward compatibility only is required. The current implementation doesn't allow this at all because the check is delegated to the [Type.checkContains](https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/schema/Type.java#L352) method.

### DIFF
--- a/parquet/src/main/scala/magnolify/parquet/Schema.scala
+++ b/parquet/src/main/scala/magnolify/parquet/Schema.scala
@@ -99,8 +99,7 @@ private object Schema {
     }
 
     writer match {
-      case _: GroupType =>
-        val wg = writer.asGroupType()
+      case wg: GroupType =>
         val rg = reader.asGroupType()
         rg.getFields.asScala.foreach { rf =>
           if (wg.containsField(rf.getName)) {
--- a/parquet/src/main/scala/magnolify/parquet/Schema.scala
+++ b/parquet/src/main/scala/magnolify/parquet/Schema.scala
@@ -20,7 +20,14 @@ import magnolify.shims.JavaConverters._
 import org.apache.parquet.io.InvalidRecordException
 import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName
 import org.apache.parquet.schema.Type.Repetition
-import org.apache.parquet.schema.{GroupType, LogicalTypeAnnotation, MessageType, Type, Types}
+import org.apache.parquet.schema.{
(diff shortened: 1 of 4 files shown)


parquet/src/main/scala/magnolify/parquet/Schema.scala
```diff
@@ -77,19 +84,28 @@ private object Schema {
     builder.named(schema.getName)
   }
 
-  def pruneRequested(fileSchema: MessageType, requestedSchema: MessageType): MessageType = {
-    def prune(file: GroupType, requested: GroupType): Type = {
-      val filtered = requested.getFields.asScala
-        .flatMap { rf =>
-          if (file.containsField(rf.getName)) {
```
clairemcginty, line 84: is there any regression now that we're no longer pruning schema fields that don't exist in `fileSchema` ? do we need to add that back in?

clairemcginty (commented):
I think this looks good! I just had some questions (inlined) while I digest it further.

parquet/src/main/scala/magnolify/parquet/Schema.scala
stormy-ua: that's a good question. Based on my observations pruning of a reader schema isn't required i.e. it is ok to pass new optional fields to the `ParquetReader` which don't exist in the writer schema. I checked this specifically in my sandbox environment and the whole schema evolution and projection suits aren't indicating any issues with removing the pruning. And it makes sense in general Parquet lib supports non-existing nullable fields in general, they wouldn't be involved into the decoding process at all i.e. converters wouldn't be created for them.

So basically we are replacing pruning with the `checkCompatibility` function.

clairemcginty (approved):

PR #455, opened 2021-11-12
add scala steward conf

stormy-ua:


.scala-steward.conf
```diff
@@ -0,0 +1,7 @@
+updates.ignore = [
+  { groupId = "com.google.cloud.datastore", artifactId = "datastore-v1-proto-client" },
```
clairemcginty, line 2: the `com.google.cloud.datastore`/`com.google.cloud` groups are probably unnecessary for now - the APIs tend to be less backwards-incompatible & the past we've merged them https://github.com/spotify/magnolify/pulls?q=is%3Apr+is%3Aclosed+datastore

same with `com.google.api.grpc`, it looks like - https://github.com/spotify/magnolify/pulls?q=is%3Apr+is%3Aclosed+proto-google-cloud-bigtable-v2+

I think avro/guava are the two main culprits :)

clairemcginty (approved):

PR #458, opened 2021-11-19
EnumType: get rid of enum instances serialization

stormy-ua:
This PR fixes a problem with serializing an instance of `ParqueType` with `EnumType` derived from a `sealed abstract class` without a no-arg constructor e.g.:

```scala
  sealed abstract class Person(val entryName: String)
  case object Aldrin extends Person("Aldrin")
  case object Neil extends Person("Neil")
``` 

The derived `ParquetType` for this enum would fail serialization with the following error:

```
val et = ensureSerializable(implicitly[EnumType[ADT.Person]])

java.io.InvalidClassException: magnolify.test.ADT$Aldrin$; no valid constructor
    at java.io.ObjectStreamClass$ExceptionInfo.newInvalidClassException(ObjectStreamClass.java:169)
    at java.io.ObjectStreamClass.checkDeserialize(ObjectStreamClass.java:874)
    at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2043)
    at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1573)
    at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2287)
    at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2211)
```
The issue could be reproduced only with Scala 2.12 and works fine with 2.13. I think there is a difference in the case object serialization between 2.12 and 2.13.

`ParquetType` is serialized and written to hadoop config to be broadcasted to all workers using HadoopIO. 

The fix handles a rare issue, but it blocks real users to move forward with `parquet-magnolify` and it isn't easy for them to get rid of the constructor with an argument because they don't control that code. Anyways, I think it is better to avoid serialization of enum instances and just initialize them lazily.

### DIFF
--- a/shared/src/main/scala/magnolify/shared/EnumType.scala
+++ b/shared/src/main/scala/magnolify/shared/EnumType.scala
@@ -32,9 +32,10 @@ sealed trait EnumType[T] extends Serializable { self =>
   def to(v: T): String
 
   def map(cm: CaseMapper): EnumType[T] = {
-    val m1 = values.map(v => cm.map(v) -> self.from(v)).toMap
-    val m2 = values.map(v => self.from(v) -> cm.map(v)).toMap
-    EnumType.create(name, namespace, values.map(cm.map), annotations, m1(_), m2(_))
+    val cMap = values.map(v => cm.map(v) -> v).toMap
+    EnumType.create(name, namespace, values.map(cm.map), annotations,
+      v => self.from(cMap(v)),
+      v => cm.map(self.to(v)))
   }
 }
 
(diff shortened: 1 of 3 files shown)


lofifnc:
This fixes also problems while running `ParquetTypeIO` in a sbt test:
```
[info]   java.lang.ClassCastException: cannot assign instance of java.lang.invoke.SerializedLambda to field magnolify.shared.EnumType$$anon$1.f$1 of type scala.Function1 in instance of magnolify.shared.EnumType$$anon$1
[info]   at java.io.ObjectStreamClass$FieldReflector.setObjFieldValues(ObjectStreamClass.java:2301)
[info]   at java.io.ObjectStreamClass.setObjFieldValues(ObjectStreamClass.java:1431)
[info]   at java.io.ObjectInputStream.defaultReadFields(ObjectInputStream.java:2411)
[info]   at java.io.ObjectInputStream.readSerialData(ObjectInputStream.java:2329)
[info]   at java.io.ObjectInputStream.readOrdinaryObject(ObjectInputStream.java:2187)
[info]   at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1667)
[info]   at java.io.ObjectInputStream.readArray(ObjectInputStream.java:2093)
```
Patching this would allow to test Parquet projection based on case class, without running a Scio job.

clairemcginty (approved):

Issue #500, opened 2022-03-09
Support FilterPredicate generation in magnolify-parquet


clairemcginty:
Similar to @nevillelyh's work in parquet-extra library: https://github.com/nevillelyh/parquet-extra/blob/main/parquet-avro/src/main/scala/me/lyh/parquet/avro/Predicate.scala - but for `T : ParquetType` types

PR #501, opened 2022-03-10
(#500) Support FilterPredicate gen for ParquetFields

clairemcginty:
adds support for custom `FilterPredicate` generation for types `T : ParquetField.Primitive`:

```scala
Predicate.onField[Int]("i1")(_ % 2 == 0)
```

I didn't go as far as adding macro support the way @nevillelyh did in [parquet-extra](https://github.com/nevillelyh/parquet-extra/blob/main/parquet-avro/src/main/scala/me/lyh/parquet/avro/Predicate.scala)... this implementation is a little more manual, maybe macro support (i.e `Predicate[RecordT)(_.i1 % 2 == 0)`) can be implemented further down the line. But it leverages the functionality of our Parquet typeclass to do things like convert between underlying Parquet types and the Scala projected type--for example, you can filter directly on `Instant` logical type fields, rather than having to assert on the raw millisecond value.

### DIFF
--- a/parquet/src/main/scala/magnolify/parquet/Predicate.scala
+++ b/parquet/src/main/scala/magnolify/parquet/Predicate.scala
@@ -13,6 +13,22 @@ import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName
 
 object Predicate {
 
+  /**
+   * Constructs a Parquet [[FilterPredicate]] based on a single primitive column and a user-defined
+   * function. Predicates can be logically combined using [[FilterApi.and]] and [[FilterApi.or]].
+   *
+   * Note that Optional fields should use a predicate for their lifted type: that is, create a
+   * predicate for an optional String field using [[Predicate.onField[String]], even if its
+   * projected case class defines the field as `Option[String]`. When applying predicates, Parquet
+   * will filter out optional fields with a value of `null` before applying the predicate filter.
+   *
+   * @param fieldName
+   *   the name of the primitive column. Nested fields are separated by '.'s
+   * @param filterFn
+   *   the UDF representing the desired filter
+   * @return
+   *   a FilterPredicate for use with typed Parquet
+   */
   def onField[ScalaFieldT](
     fieldName: String
   )(filterFn: ScalaFieldT => Boolean)(implicit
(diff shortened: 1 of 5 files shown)


parquet/src/main/scala/magnolify/parquet/Predicate.scala
```diff
@@ -0,0 +1,84 @@
...
+    def wrap[T](addFn: (PrimitiveConverter, T) => Unit): T => ScalaFieldT = {
+      val converter = pf.newConverter
+      value => {
+        addFn(converter.asPrimitiveConverter(), value)
+        converter.get
+      }
+    }
+
+    type PC = PrimitiveConverter
+    val toScalaT = (fieldType match {
```
clairemcginty, line 58: I'm torn about this converter implementation, because deferring to the `ParquetField#newConverter` means we're not duplicating any work here, but for some field types, like Integer, it seems wasteful to go through the process of buffering the bytes and creating a `PrimitiveConverter` instance when it can just be cast to the Scala type.

kellen: You could do this with `ScalaFieldT : ClassTag` and a check that the types are compatible for these few cases?

kellen (approved):
Looks good overall

parquet/src/main/scala/magnolify/parquet/Predicate.scala
clairemcginty: yeah that would work! I just did some benchmarking on 300gb of a Parquet dataset and see literally no performance difference between these two predicates:

```scala
val magnolifyPred = Predicate.onField[Long]("field")(_ % 2 == 0))

val manualPred = FilterApi.userDefined[java.lang.Long, UserDefinedPredicate[java.lang.Long] with Serializable](
        FilterApi.longColumn("field"),
        new UserDefinedPredicate[java.lang.Long] with Serializable {
        override def keep(value: lang.Long): Boolean = value % 2 == 0

        override def canDrop(statistics: Statistics[lang.Long]): Boolean = false

        override def inverseCanDrop(statistics: Statistics[lang.Long]): Boolean = false
      }))
```

So I guess we can leave it as is, for now anyway. I think the Converters recycle their buffers so it's not as wasteful as I thought.

parquet/src/test/scala/magnolify/parquet/test/ProjectionPredicateSuite.scala
```diff
@@ -125,6 +125,39 @@ class ProjectionPredicateSuite extends MagnolifySuite {
       eq.eqv(b.result(), expected)
     }
 
+  {
+    testPredicate[Wide](
```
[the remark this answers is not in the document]
clairemcginty: sounds good! added.

Issue #485, opened 2022-02-04
Add method to `ExampleType` to return the tf.schema


martinbomio:
When you need to work with tf.example, especially when you are writing them, you often also want to store the tf.schema which describes that particular tf.example dataset. It would be useful to add a methof to `ExampleType` which returns such tf.schema. Something like:
```
sealed trait ExampleType[T] extends Converter[T, Example, Example.Builder] {
   ...
  def schema: Schema
}
```

It is important to note that the returned schema object will only contain information about the names of the fields, and their types, it cannot contain any statistics information.

PR #504, opened 2022-03-23
Derive tf.schema in ExampleType

martinbomio:
Closes https://github.com/spotify/magnolify/issues/485.

Derive the tf.schema from using `ExampleType`. This will allow the serialization of such schema along with the example data.

### DIFF
--- a/tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
+++ b/tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
@@ -122,17 +122,22 @@ object ExampleField {
     override protected def buildSchema(k: String)(cm: CaseMapper): Schema = {
       val sb = Schema
         .newBuilder()
-        .setAnnotation(getDoc(caseClass.annotations, caseClass.typeName.full))
+      getDoc(caseClass.annotations, caseClass.typeName.full).foreach(sb.setAnnotation)
+
       caseClass.parameters.foldLeft(sb) { (b, p) =>
         val s = p.typeclass.buildSchema(key(k, cm.map(p.label)))(cm)
         val annotatedFs = s.getFeatureList.asScala
           .map { f =>
-            val annotation = if (f.hasAnnotation) {
-              f.getAnnotation
+            val maybeAnnotation = if (f.hasAnnotation) {
+              Some(f.getAnnotation)
             } else {
               getDoc(p.annotations, s"${caseClass.typeName.full}#${key(k, cm.map(p.label))}")
             }
-            f.toBuilder.setAnnotation(annotation).build()
+
+            maybeAnnotation match {
+              case Some(a) => f.toBuilder.setAnnotation(a).build()
+              case None    => f
+            }
           }
         b.addAllFeature(annotatedFs.asJava)
         b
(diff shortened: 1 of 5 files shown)


tensorflow/src/test/scala/magnolify/tensorflow/test/ExampleTypeSuite.scala
```diff
@@ -130,3 +165,20 @@ case class DefaultOuter(
...
+  @doc("b doc") b: Boolean,
+  @doc("i doc") i: Int,
+  @doc("l doc") l: Long,
+  @doc("f doc") f: Float,
+  @doc("d doc") d: Double,
+  @doc("bs doc") bs: ByteString,
+  @doc("s doc") s: String,
+  @doc("ii doc") ii: Array[Int],
+  @doc("ff doc") ff: Array[Float],
+  @doc("bsbs doc") bsbs: Array[ByteString],
```
kellen, line 182: Add an `Option` value for test coverage

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -23,12 +23,18 @@ import magnolia._
...
 import org.tensorflow.proto.example._
-
-import scala.annotation.implicitNotFound
+import scala.annotation.{implicitNotFound, StaticAnnotation}
+import scala.collection.concurrent
 import scala.language.experimental.macros
 
+class doc(doc: String) extends StaticAnnotation with Serializable {
+  override def toString: String = doc
+}
```
clairemcginty, line 34:
```suggestion
class doc(msg: String) extends StaticAnnotation with Serializable {
  override def toString: String = msg
}
```

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -39,15 +45,22 @@ object ExampleType {
...
 
 sealed trait ExampleField[T] extends Serializable {
+  @transient private lazy val schemaCache: concurrent.Map[ju.UUID, Schema] =
+    concurrent.TrieMap.empty
   def get(f: Features, k: String)(cm: CaseMapper): Value[T]
   def put(f: Features.Builder, k: String, v: T)(cm: CaseMapper): Features.Builder
+
+  def schema(k: String)(cm: CaseMapper): Schema =
+    schemaCache.getOrElseUpdate(cm.uuid, schemaInternal(k)(cm))
+  protected def schemaInternal(k: String)(cm: CaseMapper): Schema
```
clairemcginty, line 63: maybe we can rename this "createSchema" or "buildSchema" so it's clearer what the function is?

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -100,6 +118,36 @@ object ExampleField {
...
+        .newBuilder()
+        .setAnnotation(getDoc(caseClass.annotations, caseClass.typeName.full))
+      caseClass.parameters.foldLeft(sb) { (b, p) =>
+        val s = p.typeclass.schemaInternal(key(k, cm.map(p.label)))(cm)
+        val annotatedFs = s.getFeatureList.asScala
+          .map { f =>
+            val annotation = if (f.hasAnnotation) {
+              f.getAnnotation
+            } else {
+              getDoc(p.annotations, s"${caseClass.typeName.full}#${key(k, cm.map(p.label))}")
```
clairemcginty, line 133: just wondering, what's the rationale for this behavior (as opposed to just not calling `.setAnnotation` if `!f.hasAnnotation`)?

tensorflow/src/test/scala/magnolify/tensorflow/test/ExampleTypeSuite.scala
martinbomio: done

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
martinbomio: makes sense, will change

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
martinbomio: not sure I understand the question. We need the getDoc because the primitive example types do not add the annotations to their schema (because the schema is on the record and not on the field?)

PR #505, opened 2022-03-23
Format tests on scalafmt command

martinbomio:
The formatting checks check if the tests are formatted, adding this setting will format them when runnin `sbt scalafmt`


martinbomio:
closing since `scalafmtAll` should do this

PR #546, opened 2022-06-23
Scala 3

### LINKED ISSUE #545: Scala3 support
`com.softwaremill.magnolia1_3` is available. Try to see if we can cross-build for scala 3

RustedBones:
Fix #545 

# Modules
- [x] avro
- [x] bigquery
- [x] bigtable
- [x] cats
- [x] datastore
- [ ] docs
- [x] guava
- [ ] jmh
- [ ] parquet
- [ ] protobuf
- [ ] refined
- [x] scalacheck
- [x] shared
- [ ] tensorflow
- [x] test
- [ ] tools

# Pending Issues
- [ ] scala 3 derived class are not serializable
- [ ] annotation support
- [ ] typelcass using extra implicits (eg `scalacheck`'s `Fallback` or `datastore`'s `KeyField`) not supported
- [x] stack overflow with enum derivation
- [x] max inline: too much inline chained calls


RustedBones:
Will split the PR into chunks for every modules. See #676 for `shared` and `test`

PR #548, opened 2022-06-24
AvroType performace improvements

RustedBones:
This PR aims to improve the derived typeclass performance for Avro

Results of `jmh` benchmarks with `jmh/jmh:run -i 20 -wi 10 -f1 -t1 magnolify.jmh.AvroBench` are ~actually not~ looking good

| branch | `avroTo` | `avroFrom` | `avroSchema` |
|-----------|------------|-----------------|---------------------|
| ~PR~       |  ~8981.511 ns/op~  | ~6675.408 ns/op~ | ~3.697 ns/op~ |
| PR       |  3967.037 ns/op  | 1718.793 ns/op | 4.016 ns/op |
| main    |  8075.481 ns/op  | 3190.532 ns/op | 3.832 ns/op |
| v0.4.8  |  9463.751 ns/op  | 3417.335 ns/op | 3.767 ns/op |

Looks like `rawConstruct` is making an extra call to `MagnoliaUtil.checkParamLengths` wich is worse that going though the field map in the schema

Edit: Using vars instead improves greatly avro conversion performances (~ 40% to 50% compared to v0.4.8)

### DIFF
--- a/avro/src/main/scala/magnolify/avro/AvroType.scala
+++ b/avro/src/main/scala/magnolify/avro/AvroType.scala
@@ -125,19 +125,26 @@ object AvroField {
 
     // We are in control of the schema
     // params and fields are at the same index
-    // prefer low level constructs instead of using names
+    // use vars instead of zipWithIndex & rawConstruct for performance
     override def from(v: GenericRecord)(cm: CaseMapper): T = {
-      val params = caseClass.parameters.zipWithIndex
-        .map { case (p, idx) => p.typeclass.fromAny(v.get(idx))(cm) }
-      caseClass.rawConstruct(params)
+      var idx = -1
+      caseClass.construct { p =>
+        idx += 1
+        p.typeclass.fromAny(v.get(idx))(cm)
+      }
     }
 
+    // We are in control of the schema
+    // params and fields are at the same index
+    // use vars instead of zipWithIndex for performance
     override def to(v: T)(cm: CaseMapper): GenericRecord = {
-      caseClass.parameters.zipWithIndex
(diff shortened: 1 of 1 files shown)


avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -64,14 +63,16 @@ sealed trait AvroField[T] extends Serializable { self =>
...
   def schema(cm: CaseMapper): Schema =
-    schemaCache.getOrElseUpdate(cm.uuid, new Schema.Parser().parse(schemaString(cm)))
+    schemaCache.getOrElseUpdate(cm.uuid, createSchema(cm))
 
-  // Convert case class default `T` to Avro Java type
-  def makeDefault(p: Option[T])(cm: CaseMapper): Option[Any] = p.map(to(_)(cm))
-  // Fallback Avro Java type
-  val fallbackDefault: Any = null
+  // Convert default `T` to Avro schema default value
+  def makeDefault(d: T)(cm: CaseMapper): Any = to(d)(cm)
```
RustedBones, line 72: This is a breaking change but should not be accessed by the users.
The signature wasn't nice, we should avoid passing the applicative  as parameter to map over it and return

kellen:
Seems like codecov would be made happy by testing `DateTimeLogicalTypeFactory`

RustedBones:
Strange. It should actually be tested [here](https://github.com/spotify/magnolify/blob/c4d34c8977c078c2ecfdd62980e36929f4fa2dce/avro/src/test/scala/magnolify/avro/test/AvroTypeSuite.scala#L198)

PR #557, opened 2022-07-04
Streamline with AvroType

RustedBones:
Bring `TableRow` structure similar to `AvroType` in #548 

- No need to fallback to string schemas. Rely on clone (this is a deep copy) to construct schemas
- Use of `set` instead of `put`:

> Sets the given field value (may be null) for the given field name. Any existing value for the field will be overwritten. It may be more slightly more efficient than put(String, Object) because it avoids accessing the field's original value.

jmh perf test are similar between versions.

Caching the schema in `ParquetType`

Fixing schema cache in `ExampleType`

### DIFF
--- a/bigquery/src/main/scala/magnolify/bigquery/TableRowType.scala
+++ b/bigquery/src/main/scala/magnolify/bigquery/TableRowType.scala
@@ -63,9 +63,9 @@ sealed trait TableRowField[T] extends Serializable {
   @transient private lazy val schemaCache: concurrent.Map[ju.UUID, TableFieldSchema] =
     concurrent.TrieMap.empty
 
-  protected def createSchema(cm: CaseMapper): TableFieldSchema
+  protected def buildSchema(cm: CaseMapper): TableFieldSchema
   def fieldSchema(cm: CaseMapper): TableFieldSchema =
-    schemaCache.getOrElseUpdate(cm.uuid, createSchema(cm))
+    schemaCache.getOrElseUpdate(cm.uuid, buildSchema(cm))
 
   def from(v: FromT)(cm: CaseMapper): T
   def to(v: T)(cm: CaseMapper): ToT
@@ -87,7 +87,7 @@ object TableRowField {
   type Typeclass[T] = TableRowField[T]
 
   def join[T](caseClass: CaseClass[Typeclass, T]): Record[T] = new Record[T] {
-    override protected def createSchema(cm: CaseMapper): TableFieldSchema = {
+    override protected def buildSchema(cm: CaseMapper): TableFieldSchema = {
       // do not use a scala wrapper in the schema, so clone() works
       val fields = new ju.ArrayList[TableFieldSchema](caseClass.parameters.size)
       caseClass.parameters.foreach { p =>
@@ -147,7 +147,7 @@ object TableRowField {
         override type FromT = trf.FromT
         override type ToT = trf.ToT
 
-        override protected def createSchema(cm: CaseMapper): TableFieldSchema =
+        override protected def buildSchema(cm: CaseMapper): TableFieldSchema =
           trf.fieldSchema(cm)
         override def from(v: FromT)(cm: CaseMapper): U = f(trf.from(v)(cm))
         override def to(v: U)(cm: CaseMapper): ToT = trf.to(g(v))(cm)
@@ -157,7 +157,7 @@ object TableRowField {
   // ////////////////////////////////////////////////
 
   private def at[T](tpe: String)(f: Any => T)(g: T => Any): TableRowField[T] = new Generic[T] {
-    override protected def createSchema(cm: CaseMapper): TableFieldSchema =
+    override protected def buildSchema(cm: CaseMapper): TableFieldSchema =
       new TableFieldSchema().setType(tpe).setMode("REQUIRED")
     override def from(v: Any)(cm: CaseMapper): T = f(v)
     override def to(v: T)(cm: CaseMapper): Any = g(v)
@@ -183,7 +183,7 @@ object TableRowField {
 
   implicit def trfOption[T](implicit f: TableRowField[T]): TableRowField[Option[T]] =
     new Aux[Option[T], f.FromT, f.ToT] {
-      override protected def createSchema(cm: CaseMapper): TableFieldSchema =
+      override protected def buildSchema(cm: CaseMapper): TableFieldSchema =
         f.fieldSchema(cm).clone().setMode("NULLABLE")
       override def from(v: f.FromT)(cm: CaseMapper): Option[T] =
         if (v == null) None else Some(f.from(v)(cm))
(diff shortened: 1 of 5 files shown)


jmh/src/test/scala/magnolify/jmh/MagnolifyBench.scala
```diff
@@ -85,7 +85,7 @@ class AvroBench {
   private val genericRecord = avroType.to(nested)
   @Benchmark def avroTo: GenericRecord = avroType.to(nested)
   @Benchmark def avroFrom: Nested = avroType.from(genericRecord)
-  @Benchmark def avroSchema: Schema = avroType.schema
```
RustedBones, line 88: The was bench marking a value

parquet/src/main/scala/magnolify/parquet/ParquetType.scala
```diff
@@ -199,7 +196,14 @@ object ParquetType {
 //////////////////////////////////////////////////
 
 sealed trait ParquetField[T] extends Serializable {
-  def schema(cm: CaseMapper): Type
+
+  @transient private lazy val schemaCache: concurrent.Map[UUID, Type] =
+    concurrent.TrieMap.empty
```
RustedBones, line 201: Caching the schema as in the other type-classes

bigquery/src/main/scala/magnolify/bigquery/TableRowType.scala
```diff
@@ -112,10 +119,7 @@ object TableRowField {
     override def to(v: T)(cm: CaseMapper): TableRow =
       caseClass.parameters.foldLeft(new TableRow) { (tr, p) =>
         val f = p.typeclass.to(p.dereference(v))(cm)
-        if (f != null) {
-          tr.put(cm.map(p.label), f)
-        }
-        tr
+        if (f == null) tr else tr.set(cm.map(p.label), f)
```
RustedBones, line 122: the java doc hints that `set` performs better than `put`

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -55,12 +55,13 @@ object ExampleType {
...
+
+  protected def buildSchema(cm: CaseMapper): Schema
+  def schema(cm: CaseMapper): Schema =
+    schemaCache.getOrElseUpdate(cm.uuid, buildSchema(cm))
+
   def get(f: Features, k: String)(cm: CaseMapper): Value[T]
   def put(f: Features.Builder, k: String, v: T)(cm: CaseMapper): Features.Builder
-
-  def schema(k: String)(cm: CaseMapper): Schema =
-    schemaCache.getOrElseUpdate(cm.uuid, buildSchema(k)(cm))
```
RustedBones, line 62: This is not correct: we leak the upper key into the schema. For primitives for instance, schema will be cached with the name of the 1st fied of this type.
Schema should not contain name information of enclosing class

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -119,30 +120,23 @@ object ExampleField {
...
 
-    override protected def buildSchema(k: String)(cm: CaseMapper): Schema = {
-      val sb = Schema
-        .newBuilder()
+    override protected def buildSchema(cm: CaseMapper): Schema = {
+      val sb = Schema.newBuilder()
       getDoc(caseClass.annotations, caseClass.typeName.full).foreach(sb.setAnnotation)
-
       caseClass.parameters.foldLeft(sb) { (b, p) =>
-        val s = p.typeclass.buildSchema(key(k, cm.map(p.label)))(cm)
```
RustedBones, line 128: we see here that we couldn't rely on the parameter typeclass cached schema

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -119,30 +120,23 @@ object ExampleField {
...
-        val annotatedFs = s.getFeatureList.asScala
-          .map { f =>
-            val maybeAnnotation = if (f.hasAnnotation) {
-              Some(f.getAnnotation)
```
kellen, line 132: Aren't you missing this case from the replacement?

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -119,30 +120,23 @@ object ExampleField {
...
-              getDoc(p.annotations, s"${caseClass.typeName.full}#${key(k, cm.map(p.label))}")
-            }
-
-            maybeAnnotation match {
-              case Some(a) => f.toBuilder.setAnnotation(a).build()
-              case None    => f
-            }
-          }
-        b.addAllFeature(annotatedFs.asJava)
+        val n = cm.map(p.label)
```
kellen, line 127: Appreciate the terseness but maybe longer variable names would be more interpretable?

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -119,30 +120,23 @@ object ExampleField {
...
+        val s = p.typeclass.schema(cm)
+        val fs = s.getFeatureList.asScala.map { f =>
+          val fb = f.toBuilder
+          val k = if (f.hasName) key(n, f.getName) else n // primitives do not have name
```
kellen: primitives don't have a name ... but you call `setName` on `k` below?

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -119,30 +120,23 @@ object ExampleField {
...
-              case Some(a) => f.toBuilder.setAnnotation(a).build()
-              case None    => f
-            }
-          }
-        b.addAllFeature(annotatedFs.asJava)
+        val n = cm.map(p.label)
+        val s = p.typeclass.schema(cm)
+        val fs = s.getFeatureList.asScala.map { f =>
+          val fb = f.toBuilder
+          val fk = if (f.hasName) key(n, f.getName) else n // primitives do not have name
```
kellen, line 131: primitives don't have `name` but you call `setName` on `fk` below?

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
```diff
@@ -119,30 +120,23 @@ object ExampleField {
...
+          val d = getDoc(p.annotations, s"${caseClass.typeName.full}#$fk")
+          fb.setName(fk)
+          if (!f.hasAnnotation) d.foreach(fb.setAnnotation)
```
kellen, line 134: This seems like the reverse of what you want to do?

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
RustedBones: I mean the primitive do not have a name in their schema. Eg you can't give a name to the `Primitive[Long]`.

If the field is a primitive, we set the field name alone: `n`.
If the field is a nested record, we prepend the name `n` to the nested one `f.getName`.

tensorflow/src/main/scala/magnolify/tensorflow/ExampleType.scala
RustedBones: In the original code we have `maybeAnnotation` which is the annotation from `f` if any, or else the annotation from `p` with `getDoc`. We then set it on the builder.

Here, instead of setting again the annotation on `f` (the builder already contains it since derived from `f`), I only add the `p` annotation if there are none from `f`

PR #566, opened 2022-07-08
Add testing for tensorflow cached field schema

RustedBones:
Relates to #557 

On `v0.4.8` with test like

```scala
ef.schema("first")
assertEquals(ef.schema("second")(CaseMapper.identity), expectedFieldSchema)
```
it would fail with

```diff
=> Diff (- obtained, + expected)
 feature {
-  name: "first.b"
+  name: "b"
   type: INT
 feature {
-  name: "first.i"
+  name: "i"
   type: INT
```
`ExampleField` schema was caching the 1st parent key.

Ensured that the field `schema` does not contain cached values from wrapping class

### DIFF
--- a/tensorflow/src/test/scala/magnolify/tensorflow/test/ExampleTypeSuite.scala
+++ b/tensorflow/src/test/scala/magnolify/tensorflow/test/ExampleTypeSuite.scala
@@ -120,36 +120,46 @@ class ExampleTypeSuite extends MagnolifySuite {
   }
 
   test("WithAnnotations") {
+    implicit val ef: ExampleField[NestedWithAnnotations] = ExampleField.gen
     implicit val et: ExampleType[WithAnnotations] = ExampleType[WithAnnotations]
 
     val expectedSchema = Schema
       .newBuilder()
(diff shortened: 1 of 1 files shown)


tensorflow/src/test/scala/magnolify/tensorflow/test/ExampleTypeSuite.scala
```diff
@@ -120,36 +120,46 @@ class ExampleTypeSuite extends MagnolifySuite {
...
+      .addFeature(feature("nested.i", FeatureType.INT, Some("other i doc")))
       .setAnnotation(annotation("Example top level doc"))
       .build()
 
-    val s = et.schema
+    assertEquals(ef.schema(CaseMapper.identity), expectedFieldSchema)
     assertEquals(et.schema, expectedSchema)
   }
 
-  private def noAnnotationFeature(name: String, t: FeatureType): Feature =
```
martinbomio, line 148: this might be personal preference so feel free to ignore it, but I think is more readable to have a dedicated method instead of passing an option argument

martinbomio (approved):

PR #604, opened 2022-09-16
Typed Parquet Annotations

shnapz:
What is done:
- Added a way to annotate case classes and fields in order to add extra metadata to parquet files
- Annotations are collected in `ParquetType[T]`'s members `fieldAnnotations: Map[String, String]` and  `typeAnnotation: Option[String]` that contain respectively 1) the top level and nested fields annotations 2) the top level class annotation
- ParquetField class was moved to a dedicated file

What is left:
- Add annotations to a parquet file on write using an approach that is yet to be finalized

### DIFF
--- a/docs/parquet.md
+++ b/docs/parquet.md
@@ -48,3 +48,14 @@ implicit val pfDecimalBinary = ParquetField.decimalBinary(20, 0)
 Among the date/time types, `DATE` maps to `java.time.LocalDate`. The other types, `TIME` and `TIMESTAMP`, map to `OffsetTime`/`LocalTime` and `Instant`/`LocalDateTime` with `isAdjustedToUTC` set to `true`/`false`. They can be in nano, micro, or milliseconds precision with `import magnolify.parquet.logical.{nanos,micros,millis}._`.
 
 Note that Parquet's official Avro support maps `REPEATED` fields to an `array` field inside a nested group. Use `import magnolify.parquet.ParquetArray.AvroCompat._` to ensure compatibility with Avro.
+
+The top level class and all fields (including nested class fields) can be annotated with `@doc` annotation. Note that nested classes annotations are ignored.
+
+```scala
+@doc("This is ignored")
+case class NestedClass(@doc("nested field annotation") i: Int)
+
+@doc("Top level annotation")
+case class TopLevelType(@doc("field annotation") pd: NestedClass, @doc("field annotation 2") i: 
+Integers)
+```
--- a/parquet/src/main/scala/magnolify/parquet/ParquetField.scala
+++ b/parquet/src/main/scala/magnolify/parquet/ParquetField.scala
@@ -1,0 +1,470 @@
+/*
+ * Copyright 2022 Spotify AB
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package magnolify.parquet
+
+import magnolia1._
+import magnolify.shared.{Converter => _, _}
+import magnolify.shims.FactoryCompat
+
+import java.nio.{ByteBuffer, ByteOrder}
+import java.time.LocalDate
+import java.util.UUID
+import org.apache.parquet.io.ParquetDecodingException
+import org.apache.parquet.io.api._
+import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName
+import org.apache.parquet.schema.Type.Repetition
+import org.apache.parquet.schema.{LogicalTypeAnnotation, Type, Types}
+
+import scala.annotation.{implicitNotFound, nowarn}
+import scala.collection.concurrent
+import scala.language.experimental.macros
+import scala.collection.compat._
+
+sealed trait ParquetField[T] extends Serializable {
+
+  @transient private lazy val schemaCache: concurrent.Map[UUID, Type] =
+    concurrent.TrieMap.empty
+
+  protected def buildSchema(cm: CaseMapper): Type
+  def schema(cm: CaseMapper): Type =
+    schemaCache.getOrElseUpdate(cm.uuid, buildSchema(cm))
+
+  val hasAvroArray: Boolean = false
+  val fieldDocs: Map[String, String]
+  val typeDoc: Option[String]
+  protected val isGroup: Boolean = false
+  protected def isEmpty(v: T): Boolean
+  def write(c: RecordConsumer, v: T)(cm: CaseMapper): Unit
+  def newConverter: TypeConverter[T]
+
+  protected def writeGroup(c: RecordConsumer, v: T)(cm: CaseMapper): Unit = {
+    if (isGroup) {
+      c.startGroup()
+    }
+    write(c, v)(cm)
+    if (isGroup) {
+      c.endGroup()
+    }
+  }
+}
+
+object ParquetField {
+  sealed trait Record[T] extends ParquetField[T] {
+    override protected val isGroup: Boolean = true
+
+    override protected def isEmpty(v: T): Boolean = false
+  }
+
+  // ////////////////////////////////////////////
+  type Typeclass[T] = ParquetField[T]
+
+  def join[T](caseClass: CaseClass[Typeclass, T]): ParquetField[T] = {
+    if (caseClass.isValueClass) {
+      val p = caseClass.parameters.head
+      val tc = p.typeclass
+      new ParquetField[T] {
+        override protected def buildSchema(cm: CaseMapper): Type = tc.buildSchema(cm)
+        override protected def isEmpty(v: T): Boolean = tc.isEmpty(p.dereference(v))
+        override def write(c: RecordConsumer, v: T)(cm: CaseMapper): Unit =
+          tc.writeGroup(c, p.dereference(v))(cm)
+        override def newConverter: TypeConverter[T] = {
+          val buffered = tc.newConverter
+            .asInstanceOf[TypeConverter.Buffered[p.PType]]
+          new TypeConverter.Delegate[p.PType, T](buffered) {
+            override def get: T = inner.get(b => caseClass.construct(_ => b.head))
+          }
+        }
+        override val fieldDocs: Map[String, String] = Map.empty
+        override val typeDoc: Option[String] = None
+      }
+    } else {
+      new Record[T] {
+        override def buildSchema(cm: CaseMapper): Type =
+          caseClass.parameters
+            .foldLeft(Types.requiredGroup()) { (g, p) =>
+              g.addField(Schema.rename(p.typeclass.schema(cm), cm.map(p.label)))
+            }
+            .named(caseClass.typeName.full)
+
+        override val hasAvroArray: Boolean = caseClass.parameters.exists(_.typeclass.hasAvroArray)
+
+        override val fieldDocs: Map[String, String] =
+          caseClass.parameters.flatMap { param =>
+            val label = param.label
+            val nestedDocs = param.typeclass.fieldDocs.map { case (k, v) =>
+              s"$label.$k" -> v
+            }
+
+            val collectedAnnValue = getDoc(
+              param.annotations,
+              s"Field ${caseClass.typeName}.$label"
+            )
+            val joinedAnnotations = collectedAnnValue match {
+              case Some(value) => nestedDocs + (label -> value)
+              case None        => nestedDocs
+            }
+            joinedAnnotations
+          }.toMap
+
+        override val typeDoc: Option[String] = getDoc(
+          caseClass.annotations,
+          s"Type ${caseClass.typeName}"
+        )
+
+        override def write(c: RecordConsumer, v: T)(cm: CaseMapper): Unit = {
+          caseClass.parameters.foreach { p =>
+            val x = p.dereference(v)
+            if (!p.typeclass.isEmpty(x)) {
+              val name = cm.map(p.label)
+              c.startField(name, p.index)
+              p.typeclass.writeGroup(c, x)(cm)
+              c.endField(name, p.index)
+            }
+          }
+        }
+
+        override def newConverter: TypeConverter[T] =
+          new GroupConverter with TypeConverter.Buffered[T] {
+            private val fieldConverters = caseClass.parameters.map(_.typeclass.newConverter)
+
+            override def isPrimitive: Boolean = false
+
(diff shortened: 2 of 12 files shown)


shnapz:
To simplify a review.
One of the added things:
```
    override val fieldAnnotations: Map[String, String] =
      caseClass.parameters.flatMap { param =>
        val label = param.label
        val nestedAnnotations = param.typeclass.fieldAnnotations.map { case (k, v) =>
          (s"$label.$k", v)
        }

        val collectedAnnValue = getDoc(
          param.annotations,
          s"Field ${caseClass.typeName}.$label"
        )
        val joinedAnnotations = collectedAnnValue match {
          case Some(value) => nestedAnnotations + (label -> value)
          case None        => nestedAnnotations
        }
        joinedAnnotations
      }.toMap

    override val typeAnnotation: Option[String] = getDoc(
      caseClass.annotations,
      s"Type ${caseClass.typeName}"
    )

  private def getDoc(annotations: Seq[Any], name: String): Option[String] = {
    val docs = annotations.collect { case d: magnolify.parquet.doc => d.toString }
    require(docs.size <= 1, s"More than one @doc annotation: $name")
    docs.headOption
  }
```

docs/parquet.md
```diff
@@ -48,3 +48,14 @@ implicit val pfDecimalBinary = ParquetField.decimalBinary(20, 0)
 Among the date/time types, `DATE` maps to `java.time.LocalDate`. The other types, `TIME` and `TIMESTAMP`, map to `OffsetTime`/`LocalTime` and `Instant`/`LocalDateTime` with `isAdjustedToUTC` set to `true`/`false`. They can be in nano, micro, or milliseconds precision with `import magnolify.parquet.logical.{nanos,micros,millis}._`.
 
 Note that Parquet's official Avro support maps `REPEATED` fields to an `array` field inside a nested group. Use `import magnolify.parquet.ParquetArray.AvroCompat._` to ensure compatibility with Avro.
+
+The top level class and all fields (including nested class fields) can be annotated with `@doc` annotation. Note that nested classes annotations are ignored.
```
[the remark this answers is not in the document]
shnapz: The reason: existing Spotify avro annotations do not support type-level annotations, only field-level:
https://backstage.spotify.net/docs/default/component/data-protection-privacy/access/schemas/avsc/#my-type-doc-annotation-isnt-being-parsed-properly

Exception: the top-level type, which annotation is describing the entire dataset.
For me the intuition is simple: Spotify's data policies are concerned about data fields, because they store actual data - either sensitive or not. But nested types are auxiliary & artificial entity. According to the link above we may skip those annotations in parquet as well

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -0,0 +1,438 @@
...
+        }
+        .named(caseClass.typeName.full)
+
+    override val hasAvroArray: Boolean = caseClass.parameters.exists(_.typeclass.hasAvroArray)
+
+    override val fieldAnnotations: Map[String, String] =
+      caseClass.parameters.flatMap { param =>
+        val label = param.label
+        val nestedAnnotations = param.typeclass.fieldAnnotations.map { case (k, v) =>
+          (s"$label.$k", v)
```
RustedBones, line 87: for map entried, prefer the arrow syntax
```suggestion
          s"$label.$k" -> v
```

docs/parquet.md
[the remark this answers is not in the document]
RustedBones: This is internal to Spotify.

As OSS library, I would consider type-level annotation as a fallback if field level annotation isn't defined.
Doing so, user can define the doc once on the nested class instead of repeating the doc on every field where the nested class is used.

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -0,0 +1,438 @@
...
+  @transient private lazy val schemaCache: concurrent.Map[UUID, Type] =
+    concurrent.TrieMap.empty
+
+  protected def buildSchema(cm: CaseMapper): Type
+  def schema(cm: CaseMapper): Type =
+    schemaCache.getOrElseUpdate(cm.uuid, buildSchema(cm))
+
+  val hasAvroArray: Boolean = false
+  val fieldAnnotations: Map[String, String]
+  val typeAnnotation: Option[String]
```
RustedBones, line 48: Annotation is quite generic. What is actually supported is only the `doc` one.
```suggestion
  val fieldDocs: Map[String, String]
  val typeDoc: Option[String]
```

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
RustedBones, line 87: Prefer arrow notations for maps
```suggestion
          s"$label.$k" -> v
```

parquet/src/main/scala/magnolify/parquet/ParquetType.scala
```diff
@@ -54,10 +54,16 @@ object ParquetArray {
   }
 }
 
+class doc(doc: String) extends StaticAnnotation with Serializable {
+  override def toString: String = doc
+}
```
RustedBones, line 59: I start thinking we should define this in the `shared` module and only create a type alias in the other packages.

docs/parquet.md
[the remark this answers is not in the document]
shnapz: @RustedBones we can do type-level annotation fallback, but  I am afraid it contradicts to the existing Avro approach, so it may confuse our Spotify developers, especially if some teams support two data types

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
shnapz: Done

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
shnapz: Done

parquet/src/main/scala/magnolify/parquet/ParquetType.scala
shnapz: Done

parquet/src/main/scala/magnolify/parquet/ParquetType.scala
```diff
@@ -58,6 +52,17 @@ sealed trait ParquetType[T] extends Serializable {
...
+    val messageSchema = Schema.message(schema)
+    val avroSchema = new AvroSchemaConverter().convert(messageSchema)
+    SchemaUtil.deepCopy(
+      avroSchema,
+      typeAnnotation,
+      path => fieldAnnotations.get(path)
+    )
+  }
+  def fieldAnnotations: Map[String, String]
+  def typeAnnotation: Option[String]
```
RustedBones, line 65: This should not be exposed at the  `ParquetType` level

parquet/src/main/scala/magnolify/parquet/ParquetType.scala
```diff
@@ -58,6 +52,17 @@ sealed trait ParquetType[T] extends Serializable {
   import ParquetType._
 
   def schema: MessageType
+  def avroSchema: AvroSchema = {
+    val messageSchema = Schema.message(schema)
```
RustedBones, line 56: Useless conversion

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
```diff
@@ -0,0 +1,108 @@
...
+ * limitations under the License.
+ */
+
+package magnolify.parquet.test.util
+
+import org.apache.avro.Schema
+import cats.implicits._
+import scala.jdk.CollectionConverters._
+
+trait AvroSchemaComparer {
```
RustedBones, line 23: This is too complex. We should validate and acknowledge expected diffs if any

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
```diff
@@ -0,0 +1,108 @@
...
+            case Some(results) => results
+            case None =>
+              tryToCompareRecords(schema1, schema2) match {
+                case Some(results) => results
+                case None =>
+                  compareOtherSchemas(schema1, schema2)
+              }
+          }
+      }
+    }
```
RustedBones, line 106: This would look way better with custom `unapply` definitions and pattern matching

RustedBones (commented):
See #617

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
shnapz: Agreed

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
shnapz: Applied

parquet/src/main/scala/magnolify/parquet/ParquetType.scala
shnapz: Done

parquet/src/test/scala/magnolify/parquet/test/ParquetTypeSuite.scala
```diff
@@ -80,6 +81,26 @@ class ParquetTypeSuite extends MagnolifySuite {
...
+
+//  test("ParquetDocWithNestedList") {
+//    val pt = ensureSerializable(ParquetType[ParquetNestedListDoc])
+//
+//    assert(pt.fieldAnnotations("pd") == "nested")
+//    assert(pt.fieldAnnotations("pd.i") == "integers")
+//    assert(pt.fieldAnnotations("pd.s") == "string")
+//    assert(pt.fieldAnnotations("i") == "integers")
+//    assert(pt.typeAnnotation.contains("Parquet with doc with nested list"))
+//  }
```
RustedBones, line 102: This should be tested on the `FieldType`

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
```diff
@@ -0,0 +1,103 @@
...
+              .flatMap { case (is1, is2) => compareSchemas(is1, is2, path) }
+          }
+        case (Array(arrayEl1), Array(arrayEl2)) =>
+          compareSchemas(arrayEl1, arrayEl2, path)
+        case (Record(schemaFields1), Record(schemaFields2)) =>
+          val fields1 = schemaFields1.map(_.name())
+          val fields2 = schemaFields2.map(_.name())
+
+          val fieldsEqualResults = check(
+            !fields1.equals(fields2),
```
RustedBones, line 51: why using java equals ?

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
```diff
@@ -0,0 +1,103 @@
...
+        s"$path 'isNullable' are different '${s1.isNullable}' != '${s2.isNullable}'"
+      ) ++ check(
+        s1.getDoc != s2.getDoc,
+        s"$path 'doc' are different '${s1.getDoc}' != '${s2.getDoc}'"
+      ) ++ check(
+        Try(s1.getNamespace != s2.getNamespace).getOrElse(false),
+        s"$path 'namespace' are different '${s1.getNamespace}' != '${s2.getNamespace}'"
+      )
+    }.toList
+    else List()
```
RustedBones, line 101: prefer `List.empty` that does not create a new instance

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
```diff
@@ -0,0 +1,103 @@
...
+        s"$path 'type' are different '${s1.getType}' != '${s2.getType}'"
+      ) ++ check(
+        s1.isNullable != s2.isNullable,
+        s"$path 'isNullable' are different '${s1.isNullable}' != '${s2.isNullable}'"
+      ) ++ check(
+        s1.getDoc != s2.getDoc,
+        s"$path 'doc' are different '${s1.getDoc}' != '${s2.getDoc}'"
+      ) ++ check(
+        Try(s1.getNamespace != s2.getNamespace).getOrElse(false),
+        s"$path 'namespace' are different '${s1.getNamespace}' != '${s2.getNamespace}'"
```
RustedBones, line 98: If the Try is failing, we'll no be able to create the error string anyway

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
```diff
@@ -0,0 +1,103 @@
...
+              .flatMap { f =>
+                val field1 = schema1.getField(f)
+                val field2 = schema2.getField(f)
+
+                check(
+                  field1.doc() != field2.doc(),
+                  s"$path.$f field 'doc' are different '${field1.doc}' != '${field2.doc}'"
+                ) ++ check(
+                  field1.pos() != field2.pos(),
+                  s"$path.$f field 'defaultVal' are different '${field1.pos}' != '${field2.pos}'"
```
RustedBones, line 67: You are comparing field position here not default value

RustedBones (changes requested):
I'm not sure it is worth comparing the

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
shnapz: good catch

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
shnapz: Fixed

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
shnapz: `getNamespace()` is failing if the schema type is primitive. By design will return `false` if `Try(s1.getNamespace != s2.getNamespace).getOrElse(false)` is called for primitive type. This is verified

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
shnapz: Agreed

parquet/src/test/scala/magnolify/parquet/test/util/AvroSchemaComparer.scala
```diff
@@ -0,0 +1,103 @@
...
+          List.empty
+      }
+    }
+  }
+
+  private def check(condition: Boolean, error: => String): Option[String] = {
+    if (condition)
+      Some(error)
+    else None
+  }
```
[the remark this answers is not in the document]
shnapz: Agreed. Addressed the change

parquet/src/test/scala/magnolify/parquet/test/ParquetTypeSuite.scala
shnapz: That is fixed in your commit

RustedBones (approved):

PR #607, opened 2022-09-20
Stabilize Protobuf plugin settings for M1 compilation

clairemcginty:
I was unable to build `protobuf / test` on my M1 mac; `protobufGenerate` task failed with `error occurred while compiling protobuf files: Unsupported platform: protoc-3.11.4-osx-aarch_64.exe`.

this PR uses current recommended settings from https://github.com/sbt/sbt-protobuf. Also, we don't need to override `protobufRunProtoc`... in fact, doing so was preventing us from properly setting proto [version](https://github.com/sbt/sbt-protobuf/blob/d8f6d45078f403a0876a37cb0cba8da853b08c3f/src/main/scala/sbtprotobuf/ProtobufPlugin.scala#L67). Now we pick up the latest `com/google/protobuf/protoc` artifact with M1 support 👍

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -213,9 +213,7 @@ lazy val test: Project = project
       "org.scalameta" %% "munit-scalacheck" % munitVersion % Test,
       "org.typelevel" %% "cats-core" % catsVersion % Test
     ),
-    ProtobufConfig / protobufRunProtoc := (args =>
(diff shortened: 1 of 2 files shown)


RustedBones:
should we move the `sbt-protoc` as in `scio` instead ?

clairemcginty:
redundant to #614

PR #610, opened 2022-09-27
Treat value classes as such

turb:
The new Neo4J support works well, but value classes generate inner properties in Neo4J, when we would like them to be just strongly typed values.

### DIFF
--- a/neo4j/src/test/scala/magnolify/neo4j/ValueTypeSuite.scala
+++ b/neo4j/src/test/scala/magnolify/neo4j/ValueTypeSuite.scala
@@ -22,13 +22,10 @@ import magnolify.test.Simple._
 import magnolify.cats.auto._
 import magnolify.scalacheck.auto._
 import magnolify.shared.CaseMapper
-import org.neo4j.driver.Value
-import org.neo4j.driver.internal.value.{MapValue, StringValue}
 import org.scalacheck.{Arbitrary, Prop}
 
 import java.net.URI
 import scala.reflect.ClassTag
-import scala.jdk.CollectionConverters._
 
 class ValueTypeSuite extends MagnolifySuite {
 
@@ -69,6 +66,14 @@ class ValueTypeSuite extends MagnolifySuite {
     test[Custom]
   }
 
+  test("AnyVal") {
+    implicit val vt: ValueType[HasValueClass] = ValueType[HasValueClass]
+    test[HasValueClass]
+
+    val record = vt(HasValueClass(ValueClass("String")))
+    assert(record.get("vc").asString() == "String")
+  }
+
   test("LowerCamel mapping") {
     implicit val vt: ValueType[LowerCamel] = ValueType[LowerCamel](CaseMapper(_.toUpperCase))
     test[LowerCamel]
@@ -78,18 +83,5 @@ class ValueTypeSuite extends MagnolifySuite {
     val record = vt(LowerCamel.default)
     assert(!fields.map(record.get).exists(_.isNull))
     assert(!record.get("INNERFIELD").get("INNERFIRST").isNull)
-  }
-
-  test("AnyVal") {
-    implicit val vt: ValueType[HasValueClass] = ValueType[HasValueClass]
-    test[HasValueClass]
-
-    val record = vt(HasValueClass(ValueClass("String")))
-    assert(record.get("vc").asString() == "String")
-
-    val v: Value = new StringValue("Hello, world")
-    val a = new MapValue(Map("vc" -> v).asJava)
-    val c = vt.from(a)
-    assert(c == HasValueClass(ValueClass("Hello, world")))
   }
 }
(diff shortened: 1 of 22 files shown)


turb:
cc @RustedBones @clairemcginty

RustedBones:
This actually affects all modules.
I agree with @turb that `AnyVal`s should not be visible in the destination model after conversion.

neo4j/src/test/scala/magnolify/neo4j/ValueTypeSuite.scala
```diff
@@ -76,4 +79,17 @@ class ValueTypeSuite extends MagnolifySuite {
...
+    val vt: ValueType[HasValueClass] = implicitly
+    test[HasValueClass]
+
+    val record = vt(HasValueClass(ValueClass("String")))
+    assert(record.get("vc").asString() == "String")
+
+    val v: Value = new StringValue("Hello, world")
+    val a = new MapValue(Map("vc" -> v).asJava)
+    val c = vt.from(a)
+    assert(c == HasValueClass(ValueClass("Hello, world")))
```
RustedBones, line 93: I think this part is useless (it is already tested in the `test` while doing the round-trip)

neo4j/src/test/scala/magnolify/neo4j/ValueTypeSuite.scala
```diff
@@ -76,4 +79,17 @@ class ValueTypeSuite extends MagnolifySuite {
     assert(!fields.map(record.get).exists(_.isNull))
     assert(!record.get("INNERFIELD").get("INNERFIRST").isNull)
   }
+
+  test("AnyVal") {
+    val vt: ValueType[HasValueClass] = implicitly
```
RustedBones, line 84: Let's use the instance for the `test` too and trigger the macro with the `apply`
```suggestion
    implicit val vt: ValueType[HasValueClass] = ValueType[HasValueClass]
```

neo4j/src/test/scala/magnolify/neo4j/ValueTypeSuite.scala
turb: Not really, since the roundtrip will not test the absence of sub-property in the serialized version. Without the change, it will serialize has `{ "vs": { "str": "Hello, world" }  }` and deserialize to match the original, and the test will pass, even if the case class is an `AnyVal`.

RustedBones: What I mean, is that in the line above
```scala
assert(record.get("vc").asString() == "String")
```
We are checking that the value class does not show in the target model.
Since `test` checks the round trip, we know that the conversion back to scala will create the value class in the `HasValueClass`

turb:
> This actually affects all modules. I agree with @turb that `AnyVal`s should not be visible in the destination model after conversion.

Agreed, although for others it will be a breaking change. For Neo4J (and maybe some other databases) it won't really change something, since `AnyVal` where not usable at all.

neo4j/src/test/scala/magnolify/neo4j/ValueTypeSuite.scala
turb: Oh, yeah, the last `assert` may be seen as redundant with `test` + the previous assert. I would be in favor of testing this explicitely, however that's your call.

RustedBones:
Other modules actually have the same behavior, considering value classes as record.
I'd rather say that this is a bug in all cases (no one would use value classes due to that in any module)

turb:
@RustedBones does it mean the PR will have to wait until all cases support it?

RustedBones:
I'm preparing an update for that. I'm almost done. Would it be fine to continue on your branch?

turb:
> I'm preparing an update for that. I'm almost done. Would it be fine to continue on your branch?

Sure!

RustedBones:
With this change, `magnolia` macro do not guarantee to return a record (value-class is made transparent in the target model).
We'll have to relax the typeclass creation from the implicit, and check at runtime if it is a record

parquet/src/main/scala/magnolify/parquet/ParquetType.scala
```diff
@@ -85,19 +85,23 @@ sealed trait ParquetType[T] extends Serializable {
...
-  )(implicit f: ParquetField.Record[T], pa: ParquetArray): ParquetType[T] =
-    new ParquetType[T] {
-      override def schema: MessageType = Schema.message(f.schema(cm))
-
-      override val avroCompat: Boolean = pa == ParquetArray.AvroCompat.avroCompat || f.hasAvroArray
-      override def write(c: RecordConsumer, v: T): Unit = f.write(c, v)(cm)
-      override def newConverter: TypeConverter[T] = f.newConverter
-    }
+  )(implicit f: ParquetField[T], pa: ParquetArray): ParquetType[T] = f match {
+    case r: ParquetField.Record[_] =>
```
shnapz, line 94: Was it possible to still have a Record as implicit parameter?

RustedBones: yes, the `magnolia` macro was made to always return a `Record`

turb:
Note: Scala 2.13.9 broke binary compatibility on case classes that are value classes, so any update on it should go directly from Scala 2.13.8 to 2.13.10. See https://github.com/scala/scala/pull/10155

RustedBones:
Thanks for the heads-up @turb . Next PR #615 will go to `2.13.10` directly

PR #614, opened 2022-10-11
Setup sbt-protoc instead of sbt-protobuf

RustedBones:


build.sbt
```diff
@@ -421,23 +431,14 @@ lazy val tensorflow: Project = project
...
       "org.tensorflow" % "tensorflow-core-api" % tensorflowVersion % Provided
-    ),
-    // Protobuf plugin adds protobuf-java to Compile scope automatically; we want it to remain Provided
-    libraryDependencies := libraryDependencies.value.map { l =>
-      (l.organization, l.name) match {
-        case ("com.google.protobuf", "protobuf-java") =>
-          l.withConfigurations(Some("provided,protobuf"))
-        case _ => l
-      }
-    }
```
RustedBones, line 432: Looks this was set on the wrong project

build.sbt
```diff
@@ -409,26 +421,34 @@ lazy val protobuf: Project = project
...
-      (l.organization, l.name) match {
-        case ("com.google.protobuf", "protobuf-java") =>
-          l.withConfigurations(Some("provided,protobuf"))
-        case _ => l
-      }
+    // tensorflow metadata protos are not packaged into a jar. Manually extract them as external
+    unpackMetadata := {
+      IO.unzipURL(
+        new URL(
+          s"https://github.com/tensorflow/metadata/archive/refs/tags/v$tensorflowMetadataVersion.zip"
```
clairemcginty, line 441: this task is a great idea! 👍 

CCing @martinbomio on this change as he's contributed to magnolify-tensorflow :)

clairemcginty (approved):

build.sbt
martinbomio: would it be possible to inspect the proto files locally?
It is also important to note that this zip contains a few more proto messages than the ones strictly needed

shnapz:
I wonder if this change also solves compile errors on M1 cpus.

clairemcginty:
@shnapz just checked, it does for me--I can close #607 in favor of this one!

build.sbt
RustedBones: All files will be fetched in `tensorflow/target/protobuf_external_src/tensorflow_metadata/proto/v0`.
The zip indeed contains more files than required but they are'nt package anyway.
I'd just keep them to consider the zip as a lib, where it is common to not used every single class inside it

PR #616, opened 2022-10-12
Cleanup tests

RustedBones:
Cache the type-class test instances for `Arbitrary`, `Eq` and `EnumType` to avoid macro generation in every test suite.
Flatten test structure


avro/src/test/scala/magnolify/avro/test/AvroTypeSuite.scala
```diff
@@ -98,42 +97,37 @@ class AvroTypeSuite extends MagnolifySuite {
     }
   }
 
+  import magnolify.scalacheck.test.TestArbitrary._
```
clairemcginty, line 100: jw, why not put these in the top-level import block?

avro/src/test/scala/magnolify/avro/test/AvroTypeSuite.scala
```diff
@@ -352,8 +321,8 @@ case class MapPrimitive(m: Map[String, Int])
 case class MapNested(m: Map[String, Nested])
 
 case class Logical(u: UUID, d: LocalDate)
-case class LogicalMicros(bd: BigDecimal, i: Instant, t: LocalTime, dt: LocalDateTime)
-case class LogicalMillis(bd: BigDecimal, i: Instant, t: LocalTime, dt: LocalDateTime)
+case class LogicalMicros(i: Instant, t: LocalTime, dt: LocalDateTime)
```
clairemcginty, line 324: 👍 good catch

cats/src/test/scala/magnolify/cats/test/TestEq.scala
```diff
@@ -0,0 +1,81 @@
...
+import magnolify.test.JavaEnums
+import magnolify.test.Simple._
+import magnolify.shared.UnsafeEnum
+
+import java.net.URI
+import java.time._
+
+object TestEq {
+
+  // other
```
clairemcginty, line 31: I think this comment doesn't add clarity/context so early in the file :D maybe we should move the "simple" section above this "other" section so that "other" makes more sense in comparison

scalacheck/src/test/scala/magnolify/scalacheck/test/TestArbitrary.scala
```diff
@@ -0,0 +1,86 @@
...
+import magnolify.test.ADT._
+import magnolify.test.JavaEnums
+import magnolify.test.Simple._
+import org.scalacheck._
+
+import java.time._
+import java.net.URI
+
+object TestArbitrary {
+  // other
```
clairemcginty, line 30: same comment as above, move "simple" section up top?

clairemcginty (approved):
looks great! i have only a minor nitpick 🎨

avro/src/test/scala/magnolify/avro/test/AvroTypeSuite.scala
RustedBones: I wanted to group the implicts for/under test here but can be moved in the top level

scalacheck/src/test/scala/magnolify/scalacheck/CogenDerivationSuite.scala
```diff
@@ -45,18 +46,16 @@ class CogenDerivationSuite extends MagnolifySuite {
     }
   }
 
+  import magnolify.scalacheck.TestArbitrary._
+  implicit val cogenUri: Cogen[URI] = Cogen(_.hashCode().toLong)
```
shnapz, line 50: Why you don't use newly implemented `TestCogen.scala`?

RustedBones: Because it is the `CogenDerivationSuite` and I want to check the macros here

clairemcginty (approved):

PR #619, opened 2022-10-17
Make schema val @transient in ParquetType

clairemcginty:
There were some issues downstream in Scio because `ParquetType` couldn't be serialized as part of a `PTransform`.

Added `@transient` -- this matches the pattern we use elsewhere in Magnolify like [AvroType](https://github.com/spotify/magnolify/blob/main/avro/src/main/scala/magnolify/avro/AvroType.scala#L50), ExampleType, TableRowType etc :)

### DIFF
--- a/parquet/src/test/scala/magnolify/parquet/ParquetTypeSuite.scala
+++ b/parquet/src/test/scala/magnolify/parquet/ParquetTypeSuite.scala
@@ -44,7 +44,11 @@ class ParquetTypeSuite extends MagnolifySuite {
     t: ParquetType[T],
     eq: Eq[T]
   ): Unit = {
+    // Ensure serializable even after evaluation of `schema`
+    val parquetSchema = t.schema
+    val avroSchema = t.avroSchema
     val tpe = ensureSerializable(t)
+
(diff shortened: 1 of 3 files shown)


parquet/src/test/scala/magnolify/parquet/ParquetTypeSuite.scala
```diff
@@ -44,7 +44,11 @@ class ParquetTypeSuite extends MagnolifySuite {
     t: ParquetType[T],
     eq: Eq[T]
   ): Unit = {
+    // Ensure serializable even after evaluation of `schema`
+    val parquetSchema = t.schema
+    val avroSchema = t.avroSchema
     val tpe = ensureSerializable(t)
+
```
shnapz, line 51: Does it reproduce here - the test fail without the fix?

clairemcginty: yes, the test failed before my code change :)

shnapz (approved):

RustedBones (approved):

PR #633, opened 2022-10-26
Make parquet module compatible with Avro 1.8 and 1.11

shnapz:
Main problem: Scio pipeline fails on:
```
if (schema.isUnion) Some(schema.getTypes.asScala.toSeq) else None
```
because `isUnion` is not defined in Scio's avro version.

Scio depends on apache avro 1.8.2 and if it is increased to the latest it fails like [tests in this Scio PR](https://github.com/spotify/scio/pull/4576)
Apparently Beam really depends on somthing

Downgraded Avro in mognolify that required some changes in code as well

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,6 +51,7 @@ lazy val keepExistingHeader =
         .trim()
   )
 
+ThisBuild / versionPolicyIntention := Compatibility.BinaryAndSourceCompatible
 ThisBuild / tpolecatDefaultOptionsMode := DevMode
 ThisBuild / tpolecatDevModeOptions ~= { opts =>
   val excludes = Set(
--- a/build.sbt
+++ b/build.sbt
@@ -22,7 +22,7 @@ val magnoliaScala2Version = "1.1.2"
 val magnoliaScala3Version = "1.1.4"
 
 val algebirdVersion = "0.13.9"
-val avroVersion = Option(sys.props("avro.version")).getOrElse("1.8.2")
+val avroVersion = Option(sys.props("avro.version")).getOrElse("1.11.0")
 val bigqueryVersion = "v2-rev20220924-2.0.0"
 val bigtableVersion = "2.14.1"
 val catsVersion = "2.8.0"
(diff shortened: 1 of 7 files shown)


build.sbt
```diff
@@ -22,7 +22,7 @@ val magnoliaScala2Version = "1.1.2"
 val magnoliaScala3Version = "1.1.4"
 
 val algebirdVersion = "0.13.9"
-val avroVersion = Option(sys.props("avro.version")).getOrElse("1.11.0")
+val avroVersion = Option(sys.props("avro.version")).getOrElse("1.8.2")
```
RustedBones, line 25: Keep latest version here

build.sbt
```diff
@@ -404,7 +404,12 @@ lazy val parquet: Project = project
     description := "Magnolia add-on for Apache Parquet",
     libraryDependencies ++= Seq(
       "org.apache.hadoop" % "hadoop-client" % hadoopVersion % Provided,
-      "org.apache.parquet" % "parquet-avro" % parquetVersion % Provided,
+      "org.apache.parquet" % "parquet-avro" % parquetVersion % Provided excludeAll (
+        // parquet-avro depends on avro 1.10.x
+        ExclusionRule("org.apache.avro", "avro"),
+        ExclusionRule("org.apache.avro", "avro-compiler")
```
RustedBones, line 410: prefer `dependencyOverrides` in this case since we do not exclude but force version

RustedBones (changes requested):
The CI [avro](https://github.com/spotify/magnolify/actions/runs/3333350189/jobs/5515183354) is taking care of testing with 1.8.2. Keep normal code on the latest version. Non backward compatibility should be caught by this build

build.sbt
shnapz: When using the latest Avro - mangolify is fine. It will fail in users' app, because:
1. Scio enforces "1.8.2"
2. Even if we increase Scio version - [it will fail in Beam](https://github.com/spotify/scio/pull/4576)

build.sbt
shnapz: Done

build.sbt
shnapz: Now I got your point

RustedBones: Yes, we should only use APIs which are present in all versions.
I'm wondering if we should probably do the opposite: develop with 1.8.2 and test with 1.11.0 to make there the API is still there

shnapz: Yes, I found how to do it. Please check!

clairemcginty (approved):

RustedBones (approved):
Nice. Thanks for this one!

PR #635, opened 2022-10-27
Added versionScheme and versionPolicyCheck

shnapz:
- Added `versionPolicyCheck` to verify if our change respect the declared `versionPolicyIntention`. I am not sure on 100% the mechanism how it works in CI - it may download the previous version, but it needs to save previous reports to do it
- versionScheme := Some("early-semver") - declares our versioning strategy

https://docs.scala-lang.org/overviews/core/binary-compatibility-for-library-authors.html#versioning-scheme---communicating-compatibility-breakages

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -51,6 +51,9 @@ lazy val keepExistingHeader =
         .trim()
   )
 
+ThisBuild / versionScheme := Some("early-semver")
+ThisBuild / versionPolicyIntention := Compatibility.BinaryAndSourceCompatible
+
 ThisBuild / tpolecatDefaultOptionsMode := DevMode
 ThisBuild / tpolecatDevModeOptions ~= { opts =>
   val excludes = Set(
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -51,6 +51,9 @@ lazy val keepExistingHeader =
         .trim()
   )
 
+ThisBuild / versionScheme := Some("early-semver")
+ThisBuild / versionPolicyIntention := Compatibility.BinaryAndSourceCompatible
```
RustedBones, line 55: Is this doing the MiMa check like we do in scio ?

shnapz: Yes, plugin `"ch.epfl.scala" % "sbt-version-policy" % "2.1.0"` does MiMa check

RustedBones (changes requested):
According to the  [doc](https://github.com/scalacenter/sbt-version-policy#how-to-integrate-with-sbt-dynver), I think we need to add this too
```sbt
// Ignore dependencies to internal modules whose version is like `1.2.3+4...`
ThisBuild / versionPolicyIgnoredInternalDependencyVersions := Some("^\\d+\\.\\d+\\.\\d+\\+\\d+".r)
```

RustedBones:
I think this isn't relevant anymore since #671

PR #644, opened 2022-11-07
Do not generate faulty EnumType

### LINKED ISSUE #634: Enum implicit taken instead of typeclass derivation
In some cases, magnolia prefers deriving the `EnumType[T]` and the `enum[T: EnumType]: Typeclass[T]` instead of picking the `gen[T]: Typeclass[T]` direct derivation or scopred implicit.
This leads to the following error (on a case class containing an option):
```
initializationError  0.0s java.lang.IllegalArgumentException: 
  requirement failed: Cannot derive EnumType[T] for case class scala.Some
```
As a workaround, we can force the proper typeclass generation, but this should not happen

RustedBones:
Fix #634 

Before the PR, it was possible to generate a `EnmType[Option[ADT.Color]]` which will fail at runtime and could also mess-up with other implicit derivation, favoring enum instead of magnolia product.


RustedBones:
Do we want to go with this? @shnapz & @clairemcginty
We could potentially use the same technique to ensure that our typeclass is created on a `Product` type

RustedBones:
The other advantage of this change:
implicit `gen` and `tcEnum` do not conflict anymore:

- `gen` is defined for products. the macro fails on sealed traits due to missing `Dispatchable` implicit
- `tcEnum` expects the `EnumType` implicit that now can only be created for eums ADT.

Both macros are exclusive and won't create conflicting implicits.

Getting rid of the `shapeless.LowPriority` will also allow users to call the `tcEnum` explicitly. It was before impossible because the `shapeless.LowPriority`  parameter was expected and impossible to build manually.

RustedBones:
@shnapz what is this last commit in this branch ?
I'm not sure it has smth to do with the PR.

shared/src/main/scala/magnolify/shared/AnnotationType.scala
```diff
@@ -21,30 +21,51 @@ import scala.reflect.macros._
...
+          t <:< typeOf[scala.None.type] ||
+          t <:< typeOf[scala.Either[_, _]] ||
+          t <:< typeOf[scala.Right[_, _]] ||
+          t <:< typeOf[scala.Left[_, _]]
+      )
+
+    def impl[T: c.WeakTypeTag]: c.Tree = {
+      val wtt = weakTypeTag[T]
+      if (isStandardType(wtt.tpe)) {
+        c.abort(c.enclosingPosition, "Annotated type must not be a standard type")
```
RustedBones, line 46: Can we even annotate a type we do not declare ?

kellen (approved):

PR #647, opened 2022-11-14
Cats typeclass priority hierarchy

RustedBones:
Respect cats typeclass hierarchy for implicit resolution as defined [here](https://typelevel.org/cats/guidelines.html#implicit-instance-priority)

In practice, this was not creating any issue due to the usage of  `shapeless.LowPriority` but intelliJ linter was showing conflicts in resolution.

This fixes the lint errors and with scala 3, we'll have to define implicit priority properly anyway.
Tests were not run on the desired instance


cats/src/test/scala/magnolify/cats/ScopeTest.scala
```diff
@@ -51,3 +52,47 @@ object ScopeTest {
     ShowDerivation[Numbers]
   }
 }
+
+class ScopeTest extends FunSuite {
+
+  test("auto implicit will give most powerful abstraction") {
```
RustedBones, line 58: This is how scala solves implicit conflicts.
In our case we could have only defined those as the other implicits methods can't be called explicitly due to the shapeless `LowPriority`

cats/src/main/scala/magnolify/cats/auto/package.scala
```diff
@@ -16,4 +16,59 @@
...
 
-package object auto extends LowPriorityImplicits
+import cats._
+import cats.kernel.{Band, CommutativeGroup, CommutativeMonoid, CommutativeSemigroup}
+
+// set implicit priority to avoid conflicts
+// see: https://typelevel.org/cats/guidelines.html#implicit-instance-priority
+// use shapeless.LowPriority so the
+// provided cats type classes are always preferred
+// triggers derivation as last resort
```
RustedBones, line 26: we unfortunately have to keep `shapeless.LowPriority` to avoid conflicts with implicits provided in `cats._`

cats/src/main/scala/magnolify/cats/auto/package.scala
```diff
@@ -16,4 +16,59 @@
...
+package object auto extends LowPriority0Implicits
+
+trait LowPriority0Implicits extends LowPriority2Implicits {
```
shnapz, line 29: I wonder how this worked before the fix. Wrong type class was picked up by compiler?

cats/src/test/scala/magnolify/cats/ScopeTest.scala
```diff
@@ -19,22 +19,23 @@ package magnolify.cats
...
-    implicitly[Hash[Numbers]]
-    implicitly[Semigroup[Numbers]]
-    implicitly[CommutativeSemigroup[Numbers]]
-    implicitly[Band[Sets]]
-    implicitly[Monoid[Numbers]]
-    implicitly[CommutativeMonoid[Numbers]]
-    implicitly[Group[Numbers]]
-    implicitly[CommutativeGroup[Numbers]]
-    implicitly[Show[Numbers]]
+    val s: Show[Numbers] = implicitly
```
shnapz, line 29: is this a matter of style?

cats/src/test/scala/magnolify/cats/ScopeTest.scala
```diff
@@ -51,3 +52,47 @@ object ScopeTest {
...
   }
 }
+
+class ScopeTest extends FunSuite {
+
+  test("auto implicit will give most powerful abstraction") {
+    assertEquals(ScopeTest.Auto.s.getClass.getName, "cats.Show$$anon$2")
+    assertEquals(
+      ScopeTest.Auto.eq.getClass.getName,
+      "magnolify.cats.semiauto.HashDerivation$$anon$1"
```
shnapz, line 62: maybe it's worth to have some helper assertion syntax?

cats/src/main/scala/magnolify/cats/auto/package.scala
RustedBones: I think the `shapeless.LowPriority` got our back and was giving one of the implicit depending on the resolution.

shnapz (approved):

PR #676, opened 2023-01-18
Configure shared and test modules to cross-build scala 3

RustedBones:
cross build scala 3 test and shared modules

Some breaking changes:
- Derivation must me moved to dedicated class in specific scala folder
- All type-classes created with magnolia are created with `gen` under the type-class companion object
  - `AnnotationType.gen[T]: Annotationtype[T]`
  - `EnumType.gen[T]: EnumType[T]`
  
Thos should probably be merged into a `v0.7.x` branch to get started

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,9 +28,11 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.13.10, 2.12.17]
+        scala: [3.2.1, 2.13.10, 2.12.17]
         java: [corretto@11, corretto@8]
         exclude:
+          - scala: 3.2.1
+            java: corretto@8
           - scala: 2.12.17
             java: corretto@8
     runs-on: ${{ matrix.os }}
@@ -173,6 +175,16 @@ jobs:
             ~/AppData/Local/Coursier/Cache/v1
             ~/Library/Caches/Coursier/v1
           key: ${{ runner.os }}-sbt-cache-v2-${{ hashFiles('**/*.sbt') }}-${{ hashFiles('project/build.properties') }}
+
+      - name: Download target directories (3.2.1)
+        uses: actions/download-artifact@v3
+        with:
+          name: target-${{ matrix.os }}-${{ matrix.java }}-3.2.1
+
+      - name: Inflate target directories (3.2.1)
+        run: |
+          tar xf targets.tar
+          rm targets.tar
 
       - name: Download target directories (2.13.10)
         uses: actions/download-artifact@v3
--- a/build.sbt
+++ b/build.sbt
@@ -100,18 +100,18 @@ ThisBuild / developers := List(
 val scala3 = "3.2.1"
 val scala213 = "2.13.10"
 val scala212 = "2.12.17"
-val defaultScala = scala213
+val scalaDefault = scala213
 
 // github actions
 val java11 = JavaSpec.corretto("11")
 val java8 = JavaSpec.corretto("8")
-val defaultJava = java11
+val javaDefault = java11
 val coverageCond = Seq(
-  s"matrix.scala == '$defaultScala'",
-  s"matrix.java == '${defaultJava.render}'"
+  s"matrix.scala == '$scalaDefault'",
+  s"matrix.java == '${javaDefault.render}'"
 ).mkString(" && ")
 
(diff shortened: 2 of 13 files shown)


.github/workflows/ci.yml
```diff
@@ -28,9 +28,11 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        scala: [2.13.10, 2.12.17]
+        scala: [3.2.1, 2.13.10, 2.12.17]
         java: [corretto@11, corretto@8]
         exclude:
+          - scala: 3.2.1
```
shnapz, line 34: looks like this is no longer needed, after Java 8 deprecation PR is merged

build.sbt
```diff
@@ -186,14 +186,15 @@ val commonSettings = Seq(
...
   scalacOptions ++= {
     CrossVersion.partialVersion(scalaVersion.value) match {
       case Some((3, _)) =>
         Seq(
           // required by magnolia for accessing default values
-          "-Xretain-trees",
+          "-Yretain-trees",
           // tolerate some nested macro expansion
-          "-Ymax-inlines",
+          "-Xmax-inlines",
```
shnapz, line 197: Is it related to new Magnolia version?

build.sbt
```diff
@@ -186,14 +186,15 @@ val commonSettings = Seq(
   tlFatalWarningsInCi := false,
   tlJdkRelease := Some(8),
   tlSkipIrrelevantScalas := true,
+  scalacOptions ~= { _.filterNot(_ == "-source:3.0-migration") },
```
shnapz, line 189: can you explain what this option means?

build.sbt
```diff
@@ -296,6 +297,8 @@ lazy val scalacheck = project
     commonSettings,
     moduleName := "magnolify-scalacheck",
     description := "Magnolia add-on for ScalaCheck",
+    crossScalaVersions := Seq(scala213, scala212),
```
shnapz, line 300: so we still don't want to cross-build with scala3?

shared/src/main/scala-2/magnolify/shared/AnnotationTypeMacros.scala
```diff
@@ -0,0 +1,50 @@
+/*
```
shnapz, line 1: all code in `scala-2` was just copied?

shared/src/main/scala-3/magnolify/shims/package.scala
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package magnolify
+
+import scala.util.hashing.MurmurHash3
+import scala.collection.compat.Factory
```
shnapz, line 20: is it really needed in scala3? I thought it is only 2.12 .vs 2.13 issue

.github/workflows/ci.yml
RustedBones: This file is now generated from `build.sbt` and CI fails if it detects it is not in sync.
When we'll update the branch with changes in main, we'll just need to run `sbt githubWorkflowGenerate` to have this updated.

build.sbt
RustedBones: `source:3.0-migration` instructs the scala 3 compiler to run in [migration mode](https://docs.scala-lang.org/scala3/guides/migration/tooling-migration-mode.html). This is the default from `sbt-typlevel` when cross-building with scala 2 (the sources are the same).
However, with this option, the compiler fails on the dedicated `scala-3` folder because sources are not in the scala 2 style. See discussion https://github.com/lampepfl/dotty/discussions/16720

build.sbt
RustedBones: yes, those are required by magnolia with scala 3. Some macros are generating too long inlines.

shared/src/main/scala-3/magnolify/shims/package.scala
RustedBones: scala 2.13 and scala 3 have the same collection IIRC, but we still need the shim layer to get the code cross-build with scala 2.12. I think this can be removed when we stop 2.12 support.

clairemcginty (approved):

PR #706, opened 2023-03-07
Convert from GenericRecord using field name

shnapz:
When order of fields in case class does not match the order of fields in Avro file AvroType.from will fail
Reverting it to the way it was with previous magnolia


shnapz:
This is expected behavior

PR #729, opened 2023-04-20
Use `putUnencodedChars` for Funneling strings.

LeifW:
As per the warning on [PrimitiveSink.putString](https://javadoc.io/doc/com.google.guava/guava/latest/com/google/common/hash/PrimitiveSink.html#putString(java.lang.CharSequence,java.nio.charset.Charset)) JavaDoc:
> Warning: This method, which reencodes the input before processing it, is useful only for cross-language compatibility. For other use cases, prefer putUnencodedChars(java.lang.CharSequence), which is faster, produces the same output across Java releases, and processes every char in the input, even if some are invalid.

Or maybe leave this out and let people make their own decision as to what they want for the `Funnel[String]` instance?

### DIFF
--- a/guava/src/main/scala/magnolify/guava/semiauto/FunnelDerivation.scala
+++ b/guava/src/main/scala/magnolify/guava/semiauto/FunnelDerivation.scala
@@ -20,6 +20,8 @@ import com.google.common.base.Charsets
 import com.google.common.hash.{Funnel, Funnels, PrimitiveSink}
 import magnolia1._
 
+import scala.annotation.nowarn
+
 object FunnelDerivation {
   type Typeclass[T] = Funnel[T]
 
@@ -29,7 +31,7 @@ object FunnelDerivation {
         val p = caseClass.parameters.head
         p.typeclass.funnel(p.dereference(from), into)
       } else if (caseClass.parameters.isEmpty) {
-        into.putString(caseClass.typeName.short, Charsets.UTF_8)
+        into.putString(caseClass.typeName.short, Charsets.UTF_8): @nowarn
       } else {
         caseClass.parameters.foreach { p =>
           // inject index to distinguish cases like `(Some(1), None)` and `(None, Some(1))`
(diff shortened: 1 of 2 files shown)


guava/src/main/scala/magnolify/guava/semiauto/FunnelDerivation.scala
```diff
@@ -62,7 +62,7 @@ trait FunnelImplicits {
   implicit val charSequenceFunnel: Funnel[CharSequence] = Funnels.unencodedCharsFunnel()
 
   implicit val booleanFunnel: Funnel[Boolean] = funnel[Boolean](_.putBoolean(_))
-  implicit val stringFunnel: Funnel[String] = funnel[String](_.putString(_, Charsets.UTF_8))
+  implicit val stringFunnel: Funnel[String] = Funnels.unencodedCharsFunnel.asInstanceOf[Funnel[String]]
```
LeifW, line 65: Guess instead of using the cast, this could also be `funnel[String](_. putUnencodedChars(_))`

LeifW: If we wrote our own typeclass for this (that was equivalent to `Funnel` but defined in Scala), we could use Scala's SAM support to define instances, like `implicit val stringFunnel: OurFunnel[String] = _ putUnencodedChars _`, add methods to it like `contramap` (which is what the `by` method in here seems to be?), `widen`, add a variance annotation to it, etc... which could then be converted to a Guava Funnel at use-site?

LeifW: Since `Funnel` is invariant in the type param, maybe we could define helpers like other libraries have, e.g.
```scala
def narrowFunnel[AA, A <: AA](funnel: Funnel[AA]): Funnel[A] = funnel.asInstanceOf[Funnel[A]]
```
so then `stringFunnel` could be defined as:
```scala
implicit val stringFunnel: Funnel[String] = narrowFunnel(Funnels.unencodedCharsFunnel)
```
?
And maybe even something to not need the cast on the "unboxed" funnels like Int, Long, etc...

RustedBones: I think a simple solution would be to define a single implicit def for all `CharSequence` instances 
```scala
  implicit def charSequenceFunnel[T <: CharSequence]: Funnel[T] =
    Funnels.unencodedCharsFunnel().asInstanceOf[Funnel[T]]
 ```

File: guava/src/main/scala/magnolify/guava/semiauto/FunnelDerivation.scala [hunk 2]
```diff
@@ -59,10 +61,10 @@ trait FunnelImplicits {
   implicit val intFunnel: Funnel[Int] = Funnels.integerFunnel().asInstanceOf[Funnel[Int]]
   implicit val longFunnel: Funnel[Long] = Funnels.longFunnel().asInstanceOf[Funnel[Long]]
   implicit val bytesFunnel: Funnel[Array[Byte]] = Funnels.byteArrayFunnel()
-  implicit val booleanFunnel: Funnel[Boolean] = funnel[Boolean](_.putBoolean(_))
-  implicit val byteFunnel: Funnel[Byte] = funnel[Byte](_.putByte(_))
-  implicit val charFunnel: Funnel[Char] = funnel[Char](_.putChar(_))
-  implicit val shortFunnel: Funnel[Short] = funnel[Short](_.putShort(_))
+  implicit val booleanFunnel: Funnel[Boolean] = funnel[Boolean](_.putBoolean(_): @nowarn)
```
LeifW, line 64: Are the warnings about boxing or something? Could explicit call the associated box / unbox methods  on these - e.g. `Boolean.box` / `Boolean.unbox`

RustedBones (approved):

Issue #726, opened 2023-04-17
Support ByteBuffer derivations?


clairemcginty:
have seen a decent amount of requests for `ByteBuffer` support; although mapping to `Array[Byte]` is always an option, having built-in `ByteBuffer` support could be helpful.

shnapz:
is it Coders, magnolify, or all?

Issue #612, opened 2022-10-07
Publish docs as a GH site


clairemcginty:
similar to Scio's [GH site](https://spotify.github.io/scio/).

PR #740, opened 2023-05-04
Add joda-time and ByteBuffer support in avro

RustedBones:
avro module part of #708 and #726 

Added `val afDuration: AvroField[(Long, Long, Long)]` definition (**not implicit**) for matching duration logical type spec. Part of #255

### DIFF
--- a/avro/src/main/scala/magnolify/avro/AvroType.scala
+++ b/avro/src/main/scala/magnolify/avro/AvroType.scala
@@ -16,21 +16,22 @@
 
 package magnolify.avro
 
-import java.nio.{ByteBuffer, ByteOrder}
-import java.time._
-import java.{util => ju}
 import magnolia1._
 import magnolify.shared._
 import magnolify.shims.FactoryCompat
 import org.apache.avro.generic.GenericData.EnumSymbol
 import org.apache.avro.generic._
 import org.apache.avro.{JsonProperties, LogicalType, LogicalTypes, Schema}
+import org.joda.{time => joda}
 
+import java.nio.{ByteBuffer, ByteOrder}
+import java.time._
+import java.{util => ju}
 import scala.annotation.{implicitNotFound, nowarn}
 import scala.collection.concurrent
-import scala.reflect.ClassTag
-import scala.jdk.CollectionConverters._
 import scala.collection.compat._
+import scala.jdk.CollectionConverters._
+import scala.reflect.ClassTag
 
 sealed trait AvroType[T] extends Converter[T, GenericRecord, GenericRecord] {
   val schema: Schema
@@ -320,12 +321,12 @@ object AvroField {
   // date
   implicit val afDate: AvroField[LocalDate] =
     logicalType[Int](LogicalTypes.date())(x => LocalDate.ofEpochDay(x.toLong))(_.toEpochDay.toInt)
-  private val EpochJodaDate = new org.joda.time.LocalDate(1970, 1, 1)
-  implicit val afJodaDate: AvroField[org.joda.time.LocalDate] =
+  private lazy val EpochJodaDate = new joda.LocalDate(1970, 1, 1)
+  implicit val afJodaDate: AvroField[joda.LocalDate] =
     logicalType[Int](LogicalTypes.date()) { daysFromEpoch =>
       EpochJodaDate.plusDays(daysFromEpoch)
     } { date =>
-      org.joda.time.Days.daysBetween(EpochJodaDate, date).getDays
+      joda.Days.daysBetween(EpochJodaDate, date).getDays
     }
 
   // duration, as in the avro spec. do not make implicit as there is not a specific type for it
--- a/avro/src/test/scala/magnolify/avro/AvroTypeSuite.scala
+++ b/avro/src/test/scala/magnolify/avro/AvroTypeSuite.scala
@@ -18,29 +18,27 @@ package magnolify.avro
 
 import cats._
 import com.fasterxml.jackson.core.JsonFactory
-import com.fasterxml.jackson.databind.JsonNode
-import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.{JsonNode, ObjectMapper}
 import magnolify.avro.unsafe._
-import magnolify.cats.auto._
 import magnolify.cats.TestEq._
-import magnolify.scalacheck.auto._
+import magnolify.cats.auto._
 import magnolify.scalacheck.TestArbitrary._
+import magnolify.scalacheck.auto._
 import magnolify.shared.CaseMapper
 import magnolify.shared.TestEnumType._
 import magnolify.test.Simple._
 import magnolify.test._
 import org.apache.avro.Schema
 import org.apache.avro.generic._
-import org.apache.avro.io.DecoderFactory
-import org.apache.avro.io.EncoderFactory
+import org.apache.avro.io.{DecoderFactory, EncoderFactory}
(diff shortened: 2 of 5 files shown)


avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -196,14 +197,14 @@ object AvroField {
...
   implicit val afFloat: AvroField[Float] = id[Float](Schema.Type.FLOAT)
   implicit val afDouble: AvroField[Double] = id[Double](Schema.Type.DOUBLE)
-  implicit val afBytes: AvroField[Array[Byte]] = new Aux[Array[Byte], ByteBuffer, ByteBuffer] {
+  implicit val afByteBuffer: AvroField[ByteBuffer] = new Aux[ByteBuffer, ByteBuffer, ByteBuffer] {
     override protected def buildSchema(cm: CaseMapper): Schema = Schema.create(Schema.Type.BYTES)
     // `JacksonUtils.toJson` expects `Array[Byte]` for `BYTES` defaults
-    override def makeDefault(d: Array[Byte])(cm: CaseMapper): Array[Byte] = d
-    override def from(v: ByteBuffer)(cm: CaseMapper): Array[Byte] =
-      ju.Arrays.copyOfRange(v.array(), v.position(), v.limit())
-    override def to(v: Array[Byte])(cm: CaseMapper): ByteBuffer = ByteBuffer.wrap(v)
```
RustedBones, line 205: Here we have a non symmetric definition:
- in the `from` we copy the data
- in the `to` we only wrap, meaning a change to the converted avro object will modify the scala array.

avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -196,14 +197,14 @@ object AvroField {
...
+    override def makeDefault(d: ByteBuffer)(cm: CaseMapper): Array[Byte] = d.array()
+    override def from(v: ByteBuffer)(cm: CaseMapper): ByteBuffer = v
+    override def to(v: ByteBuffer)(cm: CaseMapper): ByteBuffer = v
   }
+  implicit val afBytes: AvroField[Array[Byte]] = from[ByteBuffer](_.array())(ByteBuffer.wrap)
```
RustedBones, line 207: Changed the behavior to pass the same mutable array from avro to scala

avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -16,7 +16,7 @@
 
 package magnolify.avro
 
-import java.nio.ByteBuffer
+import java.nio.{ByteBuffer, ByteOrder}
 import java.time._
```
clairemcginty, line 20: Maybe we could import joda as `import org.joda.{time => joda}` so that the Joda APIs could simply refer to `joda.LocalDate`?

avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -196,14 +196,14 @@ object AvroField {
   implicit val afLong: AvroField[Long] = id[Long](Schema.Type.LONG)
   implicit val afFloat: AvroField[Float] = id[Float](Schema.Type.FLOAT)
   implicit val afDouble: AvroField[Double] = id[Double](Schema.Type.DOUBLE)
-  implicit val afBytes: AvroField[Array[Byte]] = new Aux[Array[Byte], ByteBuffer, ByteBuffer] {
+  implicit val afByteBuffer: AvroField[ByteBuffer] = new Aux[ByteBuffer, ByteBuffer, ByteBuffer] {
```
shnapz, line 199: looks like it could be declared as `id[ByteBuffer](Schema.Type. BYTES)`, but `id` needs to have an override to accommodate `d.array()`

avro/src/test/scala/magnolify/avro/AvroTypeSuite.scala
```diff
@@ -190,7 +197,29 @@ class AvroTypeSuite extends MagnolifySuite {
...
+      for {
+        months <- Gen.chooseNum(0L, 0xffffffffL)
+        days <- Gen.chooseNum(0L, 0xffffffffL)
+        millis <- Gen.chooseNum(0L, 0xffffffffL)
+      } yield AvroDuration(months, days, millis)
+    }
+
+    implicit val afAvroDuration: AvroField[AvroDuration] = AvroField.from[(Long, Long, Long)] {
+      case (months, days, millis) => AvroDuration(months, days, millis)
+    }(d => (d.months, d.days, d.millis))(AvroField.afDuration)
```
shnapz, line 212: We either need explain how to use `AvroField.afDuration` in docs, or introduce `AvroDuration` in magnolify. It is concerning that it won't be useful to users

avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -316,8 +316,37 @@ object AvroField {
...
+      EpochJodaDate.plusDays(daysFromEpoch)
+    } { date =>
+      org.joda.time.Days.daysBetween(EpochJodaDate, date).getDays
+    }
+
+  // duration, as in the avro spec. do not make implicit as there is not a specific type for it
+  // A duration logical type annotates Avro fixed type of size 12, which stores three little-endian unsigned integers
+  // that represent durations at different granularities of time.
+  // The first stores a number in months, the second stores a number in days, and the third stores a number in milliseconds.
+  val afDuration: AvroField[(Long, Long, Long)] =
```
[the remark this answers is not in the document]
RustedBones: `Duration` is not made to hold `month` but `Period` can, so we could define as `(Period, Duration)`.
I was unsure though where to store the `days`, as both can hold the value. I think we can split like this:
- `Period`: months + days
- `Duration`: milliseconds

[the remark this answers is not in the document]
RustedBones: Actually, `Period` is defined as (years: Int, months: Int, days: Int), so we can have an overflow when converting from avro, as the specification tells the value are unsinged integers (handled as long)

avro/src/test/scala/magnolify/avro/AvroTypeSuite.scala
RustedBones: I'm not for exposing a custom `AvroDuration` class. IHMO, the avro java lib does not expose the `duration` logical type because there does not exists a proper java data structure to handle the value as defined in the spec.

In the end `afDuration` is not different from any other type, so users should not be lost.

avro/src/main/scala/magnolify/avro/AvroType.scala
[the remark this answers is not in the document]
clairemcginty: ok, got it. This is probably the least painful solution, then -- can you add documentation (both on how to import it, and on what it represents) to https://github.com/spotify/magnolify/blob/main/docs/avro.md + https://github.com/spotify/magnolify/blob/main/docs/mapping.md ?

avro/src/test/scala/magnolify/avro/AvroTypeSuite.scala
shnapz: What I was thinking is that it might be not obvious for a user how to discover `afDuration`

shnapz (approved):

avro/src/test/scala/magnolify/avro/AvroTypeSuite.scala
RustedBones: We can include that in #612 . We'll need to create a proper doc for all modules

PR #783, opened 2023-07-05
Update to sbt-typelevel 0.5.0

RustedBones:
compliation now fails for public implicit if type is not explicitelly specified


.github/workflows/ci.yml
```diff
@@ -15,74 +15,57 @@ on:
     tags: [v*]
 
 env:
-  PGP_PASSPHRASE: ${{ secrets.PGP_PASSPHRASE }}
```
shnapz, line 18: I guess it was never used?

RustedBones: It was moved to the step envs instead of the global one

PR #788, opened 2023-07-10
Switch from abstract val to def in traits

RustedBones:
Prefer using `def` over abstract `val` in traits.
This increases implementation flexibility and avoid unnecessary serialization and memory usage (when refering to a constant for instance)


parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -42,14 +42,14 @@ sealed trait ParquetField[T] extends Serializable {
...
+  def hasAvroArray: Boolean = false
   def fieldDocs(cm: CaseMapper): Map[String, String]
   def typeDoc: Option[String]
 
-  protected val isGroup: Boolean = false
+  protected def isGroup: Boolean = false
   protected def isEmpty(v: T): Boolean
   def write(c: RecordConsumer, v: T)(cm: CaseMapper): Unit
-  def newConverter: TypeConverter[T]
+  def newConverter(): TypeConverter[T]
```
RustedBones, line 52: This function has a side effect and should hence be defined with parenthesis

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
```diff
@@ -133,7 +133,7 @@ object ProtobufField {
       new ProtobufField[T] {
         override type FromT = tc.FromT
         override type ToT = tc.ToT
-        override val hasOptional: Boolean = tc.hasOptional
+        override def hasOptional: Boolean = tc.hasOptional
```
kellen, line 136: In these implementation cases, isn't it preferable to define these fields as `val` (or maybe `lazy val`?)

kellen (approved):

protobuf/src/main/scala/magnolify/protobuf/ProtobufType.scala
RustedBones: the only case where there is some computation is in `Record` where it is already a val. IMHO here it is not worth caching another constant

PR #818, opened 2023-08-23
Support proto3 optional

### LINKED ISSUE #814: Replace deprecated usage of `FileDescriptor.Syntax` in ProtobufType
`FileDescriptor.Syntax` has been marked as deprecated.
Check what is the replacement and use it in `ProtobufType`

RustedBones:
Since [protobuf 3.15.0](https://github.com/protocolbuffers/protobuf/releases/tag/v3.15.0) (2021-02), `optional` is supported by default in the proto 3 syntax.
Adapt the `ProtobufType` typeclass to take this into account.

As side effect, this removes the need to check for proto 2 syntax.

Fix #814


kellen (approved):
When readers on older proto3 versions read these optional fields, what happens?

RustedBones:
I think the old proto3 will always return `true` to the `hasPresence` since the generated field won't be a `oneOf`. The read field will be a `Some` of the default value, as it was before.

PR #841, opened 2023-10-04
Hbase typeclass

RustedBones:
Add `hbase` module to support conversion from`Result` ant to `Put` mutation

### DIFF
--- a/hbase/src/main/java/magnolify/hbase/HBaseType.scala
+++ b/hbase/src/main/java/magnolify/hbase/HBaseType.scala
@@ -26,7 +26,7 @@ import java.nio.charset.StandardCharsets.UTF_8
 import java.util.UUID
 import scala.annotation.implicitNotFound
 import scala.jdk.CollectionConverters.*
-sealed trait HbaseType[T] extends Converter[T, Map[String, Array[Byte]], Map[String, Array[Byte]]] {
+sealed trait HBaseType[T] extends Converter[T, Map[String, Array[Byte]], Map[String, Array[Byte]]] {
   def apply(v: Result, family: Array[Byte]): T =
     from(
       v.getFamilyMap(family)
@@ -41,12 +41,12 @@ sealed trait HbaseType[T] extends Converter[T, Map[String, Array[Byte]], Map[Str
     }
 }
 
-object HbaseType {
-  implicit def apply[T: HbaseField]: HbaseType[T] = HbaseType(CaseMapper.identity)
+object HBaseType {
+  implicit def apply[T: HBaseField]: HBaseType[T] = HBaseType(CaseMapper.identity)
 
-  def apply[T](cm: CaseMapper)(implicit f: HbaseField[T]): HbaseType[T] = f match {
-    case r: HbaseField.Record[_] =>
-      new HbaseType[T] {
+  def apply[T](cm: CaseMapper)(implicit f: HBaseField[T]): HBaseType[T] = f match {
+    case r: HBaseField.Record[_] =>
+      new HBaseType[T] {
         private val caseMapper: CaseMapper = cm
 
         override def from(xs: Map[String, Array[Byte]]): T = r.get(xs, null)(caseMapper).get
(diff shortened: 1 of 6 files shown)


hbase/src/main/java/magnolify/hbase/HBaseType.scala
```diff
@@ -0,0 +1,267 @@
...
+import magnolify.shared.*
+import magnolify.shims.FactoryCompat
+import org.apache.hadoop.hbase.client.{Mutation, Put, Result}
+
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets.UTF_8
+import java.util.UUID
+import scala.annotation.implicitNotFound
+import scala.jdk.CollectionConverters.*
+sealed trait HBaseType[T] extends Converter[T, Map[String, Array[Byte]], Map[String, Array[Byte]]] {
```
clairemcginty, line 29:
```suggestion

sealed trait HBaseType[T] extends Converter[T, Map[String, Array[Byte]], Map[String, Array[Byte]]] {
```

hbase/src/main/java/magnolify/hbase/HBaseType.scala
```diff
@@ -0,0 +1,267 @@
...
+      xs.get(k) match {
+        case Some(v) => Value.Some(fromBytes(v))
+        case None    => Value.None
+      }
+
+    override def put(k: String, v: T)(cm: CaseMapper): Map[String, Array[Byte]] =
+      Map(k -> toBytes(v))
+  }
+
+  // ////////////////////////////////////////////////
```
clairemcginty, line 84: missing a comment here or is it just intended as a linebreak?

hbase/src/main/java/magnolify/hbase/HBaseType.scala
```diff
@@ -0,0 +1,267 @@
...
+          var fallback = true
+          val r = caseClass.construct { p =>
+            val q = qualifier(k, cm.map(p.label))
+            val v = p.typeclass.get(xs, q)(cm)
+            if (v.isSome) {
+              fallback = false
+            }
+            v.getOrElse(p.default)
+          }
+          // result is default if all fields are default
```
clairemcginty, line 113: Unless I'm misunderstanding the `construct` lambda, it looks like `result` defaults if even one field is default?

hbase/src/main/java/magnolify/hbase/HBaseType.scala
```diff
@@ -0,0 +1,267 @@
...
+    primitive[Int](java.lang.Integer.BYTES)(_.getInt)(_.putInt(_))
+  implicit val hbfLong: Primitive[Long] =
+    primitive[Long](java.lang.Long.BYTES)(_.getLong)(_.putLong(_))
+  implicit val hbfFloat: Primitive[Float] =
+    primitive[Float](java.lang.Float.BYTES)(_.getFloat)(_.putFloat(_))
+  implicit val hbfDouble: Primitive[Double] =
+    primitive[Double](java.lang.Double.BYTES)(_.getDouble)(_.putDouble(_))
+  implicit val hbfBoolean: Primitive[Boolean] =
+    from[Byte](_ == 1)(if (_) 1 else 0)
+  implicit val hbfUUID: Primitive[UUID] =
```
clairemcginty, line 169: nice! should we include support for java.time classes as well?

hbase/src/main/java/magnolify/hbase/HBaseType.scala
RustedBones: linebreak as in the other modules

hbase/src/main/java/magnolify/hbase/HBaseType.scala
RustedBones: Maybe in an unsafe package. As there could be several way to store dates (string bytes, ms long bytes, ...), we should not arbitrarily prick one here.

hbase/src/main/java/magnolify/hbase/HBaseType.scala
RustedBones: I copied this from the `bigtable` module. I don't get either why we use the `Value` wrapper. I'd see If I can clean that

RustedBones: See https://github.com/spotify/magnolify/pull/847

kellen:
See https://github.com/spotify/scio/pull/5010#issuecomment-2627548763

PR #847, opened 2023-10-24
Map based converter for flat model

RustedBones:
When object is not a tree, use a map of qualified keys to values as converter source.
Filtering keys when cursor selects field removes the need to wrap the return type into a Value


bigtable/src/main/scala/magnolify/bigtable/BigtableType.scala
```diff
@@ -16,26 +16,25 @@
...
 
-sealed trait BigtableType[T] extends Converter[T, java.util.List[Column], Seq[SetCell.Builder]] {
+sealed trait BigtableType[T] extends Converter[T, Map[String, Cell], Seq[SetCell.Builder]] {
   def apply(v: Row, columnFamily: String): T =
     from(
       v.getFamiliesList.asScala
         .find(_.getName == columnFamily)
-        .map(_.getColumnsList)
-        .getOrElse(java.util.Collections.emptyList())
+        .map(_.getColumnsList.asScala.map(c => c.getQualifier.toStringUtf8 -> c.getCells(0)).toMap)
```
RustedBones, line 36: we should probably keep the `Column` here instad of the cell in case we get extra colums with empty cells

clairemcginty (approved):

PR #857, opened 2023-11-08
Cross-build scalacheck module on scala 3

RustedBones:


scalacheck/src/test/scala/magnolify/scalacheck/CogenDerivationSuite.scala
```diff
@@ -16,25 +16,33 @@
...
     implicit val arbList: Arbitrary[List[T]] = Arbitrary(Gen.listOfN(10, arb.arbitrary))
     property(s"$name.uniqueness") {
       Prop.forAll { (seed: Seed, xs: List[T]) =>
-        xs.map(co.perturb(seed, _)).toSet.size == xs.toSet.size
+        val coper = xs.map(co.perturb(seed, _))
+        val result = coper.toSet.size == xs.toSet.size
+        if (!result) {
+          println("coper: " + coper)
+          println("origin: " + xs)
+        }
```
RustedBones, line 44: trace to remove

scalacheck/src/main/scala-2/magnolify/scalacheck/ArbitraryDerivation.scala
```diff
@@ -0,0 +1,58 @@
+/*
+ * Copyright 2019 Spotify AB
```
clairemcginty, line 2:
```suggestion
 * Copyright 2023 Spotify AB
```

scalacheck/src/main/scala-2/magnolify/scalacheck/ArbitraryDerivation.scala
```diff
@@ -0,0 +1,58 @@
...
+      Gen.sized { size =>
+        val subtypes = sealedTrait.subtypes
+        for {
+          i <-
+            if (size >= 0) {
+              // pick any subtype
+              Gen.choose(0, subtypes.size - 1)
+            } else {
+              // pick a fixed subtype to have a chance to stop recursion
+              Gen.const(subtypes.size + size)
```
clairemcginty, line 46: I'm not totally following the `subtypes.size + size` logic, can you explain a bit?

scalacheck/src/main/scala-2/magnolify/scalacheck/ArbitraryDerivation.scala
```diff
@@ -0,0 +1,58 @@
...
+            }
+          subtypeGen <- Gen.resize(size - 1, sealedTrait.subtypes(i).typeclass.arbitrary)
+        } yield subtypeGen
+      }
+    }
+  }
+
+  implicit def gen[T]: Arbitrary[T] = macro Magnolia.gen[T]
```
clairemcginty, line 54: I like this signature change, it matches the Scalacheck API better too imo 👍

scalacheck/src/main/scala-2/magnolify/scalacheck/ArbitraryDerivation.scala
RustedBones: Once we reach negative size, we'll stop picking type at random, and deterministically pick one of the subtype. This is done so we'll have a chance to end-up recursion when tree structure is big.

The doenside is that such generated will often have the same kind of tali/leaves

RustedBones: This was previously done by providing the `Fallback` implicit.

PR #876, opened 2023-12-18
Parquet map logical type

RustedBones:
Implementing `Map[K, V]` as described by parquet MAP logical type [documentation](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#maps)

### DIFF
--- a/parquet/src/main/scala/magnolify/parquet/ParquetField.scala
+++ b/parquet/src/main/scala/magnolify/parquet/ParquetField.scala
@@ -48,6 +48,8 @@ sealed trait ParquetField[T] extends Serializable {
 
   protected val isGroup: Boolean = false
   protected def isEmpty(v: T): Boolean
+  protected final def nonEmpty(v: T): Boolean = !isEmpty(v)
+
   def write(c: RecordConsumer, v: T)(cm: CaseMapper): Unit
   def newConverter: TypeConverter[T]
 
@@ -128,7 +130,7 @@ object ParquetField {
         override def write(c: RecordConsumer, v: T)(cm: CaseMapper): Unit = {
           caseClass.parameters.foreach { p =>
             val x = p.dereference(v)
-            if (!p.typeclass.isEmpty(x)) {
+            if (p.typeclass.nonEmpty(x)) {
               val name = cm.map(p.label)
               c.startField(name, p.index)
               p.typeclass.writeGroup(c, x)(cm)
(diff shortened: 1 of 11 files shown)


parquet/src/test/scala/magnolify/parquet/ParquetTypeSuite.scala
```diff
@@ -193,6 +196,8 @@ class ParquetTypeSuite extends MagnolifySuite {
 
 case class Unsafe(c: Char)
 case class ParquetTypes(b: Byte, s: Short, ba: Array[Byte])
+case class MapPrimitive(strMap: Map[String, Int])
```
clairemcginty, line 199: Can we also test a nullable value type, since that's allowed in the Parquet spec?

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -366,6 +366,68 @@ object ParquetField {
     }
   }
 
+  private val KeyField = "key"
+  private val ValueField = "value"
+  private val MapGroup = "key_value"
+  implicit def pfMap[K, V](implicit
+    pfKey: ParquetField[K],
```
clairemcginty, line 373: possible to validate here that `pfKey` is not a nullable type?

clairemcginty (approved):
awesome! thanks for updating `tools` as well!

parquet/src/test/scala/magnolify/parquet/ParquetTypeSuite.scala
RustedBones: gotcha. My implementation contained a bug. Thanks!

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -366,6 +368,95 @@ object ParquetField {
...
+          .addField(valueSchema)
+          .named(KeyValueGroup)
+        Types
+          .requiredGroup()
+          .addField(keyValue)
+          .as(LogicalTypeAnnotation.mapType())
+          .named("map")
+      }
+
+      override val hasAvroArray: Boolean = pfKey.hasAvroArray || pfValue.hasAvroArray
```
clairemcginty, line 395: I almost think we should throw an error if there's an `hasAvroArray` mismatch between key and value (and they both contain repeated types), since that probably points to a user error... that said, an iterable key is enough of an edge case it might not be worth implementing 😅

clairemcginty (approved):

PR #886, opened 2024-01-11
Fix mdoc wiring in site

clairemcginty:
I noticed that mdoc execution results weren't rendering in the site, even for `mdoc` code blocks that aren't marked `compile-only`, i.e. enums.md: https://github.com/spotify/magnolify/blob/main/site/src/main/paradox/enums.md?plain=1#L7

You can see on the current GH site, the mdoc rendering is not included: https://spotify.github.io/magnolify/enums.html
Screenshot 2024-01-11 at 3 20 21 PM

Looking through the target directories created by `sbt site/makeSite`, it looked like the output of `site/target/mdoc` contained the correct mdoc rendering:

```md
site/target/mdoc/enums.md:

```scala
object Color extends Enumeration {
  type Type = Value
  val Red, Green, Blue = Value
}

import magnolify.shared._
// Encode as ["red", "green", "blue"]
implicit val enumType: EnumType[Color.Type] = EnumType.scalaEnumType[Color.Type].map(CaseMapper(_.toLowerCase))
// enumType: EnumType[Color.Type] = magnolify.shared.EnumType$$anon$1@53c448db
```

but `site/target/paradox` didn't:

```md
site/target/paradox/site/main/enums.html:

```html
<p><code>CaseMapper</code> supports enums too.</p>
<pre class="prettyprint"><code class="language-scala mdoc">object Color extends Enumeration {
  type Type = Value
  val Red, Green, Blue = Value
}

import magnolify.shared._
// Encode as [&quot;red&quot;, &quot;green&quot;, &quot;blue&quot;]
implicit val enumType: EnumType[Color.Type] = EnumType.scalaEnumType[Color.Type].map(CaseMapper(_.toLowerCase))
</code></pre>
```

It looked like Paradox was basically re-using `site/src/main/paradox` [as a sourceDirectory](https://github.com/lightbend/paradox/blob/master/plugin/src/main/scala/com/lightbend/paradox/sbt/ParadoxPlugin.scala#L89-L96).

To fix, I configured `paradox / sourceDirectory` to be `mdocOut.value`, which fixed the problem:

Screenshot 2024-01-11 at 3 27 26 PM


RustedBones:
Good finding.
I'd prefer to stick with default plugin layout then:
Put sources in `doc/` folder as it used to, and add the `mdocOut` as managed input for paradox

PR #921, opened 2024-02-20
Cache TensorFlow Metadata proto files

RustedBones:
See https://github.com/spotify/scio/pull/5246

All credits to @Duhemm


clairemcginty (approved):
nice 🎉

PR #942, opened 2024-03-27
[shared] Add scala 2.12 CanBuildFrom for Set

RustedBones:
Set could be use as collection in scala 2.13 but not in 2.12 due to a missing serializable can build from factory.


Proximyst (approved):
This looks good for our team :)

kellen (approved):

PR #988, opened 2024-06-12
[protobuf] Add map support

RustedBones:
`protobuf` allows to define [maps](https://protobuf.dev/programming-guides/proto3/#maps).


rscarvalho (commented):
Thanks for adding this! One piece we might also need to update in the [type mapping table](https://github.com/spotify/magnolify/blob/main/docs/mapping.md#type-mapping) in the docs.

PR #996, opened 2024-07-09
(fix #766) Deprecate AvroCompat, replace automatic schema detection on read + Configurable write

### LINKED ISSUE #766: Rework AvroCompat for ParquetType
**Background:**
Parquet doesn’t have a single canonical in-memory representation like Avro does; it’s a file format whose read/write layer allows the user to select the specific data format they’d like to read Parquet records into. Parquet-Avro, which belongs to the OSS Parquet library, is one of the most popular. Magnolify-Parquet provides an alternative data format: Scala case classes. Theoretically, data formats can be mixed and matched; you can write using parquet-avro and read into Scala case classes, or vice versa.

The exception to data format interchangeability is when a schema contains a repeated field. Parquet’s MessageType natively supports marking any primitive or complex field as repeated, for example:

```proto
message TestRecord {
  required int64 intField (INTEGER(64,true));
  repeated binary listField (STRING);
}
```

By default, this is how Magnolify-Parquet generates schemas for repeated types. However, this protocol for repeated fields is incompatible with the Parquet-Avro format:

```scala
scala> import magnolify.parquet._
scala> case class TestRecord(intField: Int, listField: List[String])
scala> ParquetType[TestRecord].schema
val res0: org.apache.parquet.schema.MessageType =
message TestRecord {
  required int32 intField (INTEGER(32,true));
  repeated binary listField (STRING);
}
scala> ParquetType[TestRecord].avroSchema
java.lang.UnsupportedOperationException: REPEATED not supported outside LIST or MAP. Type: repeated binary listField (STRING)
  at org.apache.parquet.avro.AvroSchemaConverter.convertFields(AvroSchemaConverter.java:292)
  at org.apache.parquet.avro.AvroSchemaConverter.convert(AvroSchemaConverter.java:279)
  at magnolify.parquet.ParquetType$$anon$1.avroSchema$lzycompute(ParquetType.scala:89)
  at magnolify.parquet.ParquetType$$anon$1.avroSchema(ParquetType.scala:88)
  ... 59 elided
```

As a result, Parquet records containing a repeated field that are written with Parquet-Avro could be read using Magnolify-Parquet, until we introduced [AvroCompat](https://github.com/spotify/magnolify/blob/a8c5b6710b1a395e1529b35154f98f098d3bbaf2/parquet/src/main/scala/magnolify/parquet/ParquetType.scala#L44) to Magnolify. This was a Scala object that, when imported, used an encoding trick to produce repeated schemas compatible with Parquet-Avro:

```scala
scala> import magnolify.parquet.ParquetArray.AvroCompat._
scala> ParquetType[TestRecord].schema
val res1: org.apache.parquet.schema.MessageType =
message TestRecord {
  required int32 intField (INTEGER(32,true));
  required group listField (LIST) {
    repeated binary array (STRING);
  }
}
scala> ParquetType[TestRecord].avroSchema
val res2: org.apache.avro.Schema = {"type":"record","name":"TestRecord","namespace":"$iw","fields":[{"name":"intField","type":"int"},{"name":"listField","type":{"type":"array","items":"string"}}]}
```

**Problem Statement:**
This works, but has some downsides that have become more apparent with increased adoption, namely:

1. The mechanism itself (importing a class) is not intuitive to users and easy to forget to add – and once you’ve started producing data with a certain schema, it’s non-trivial to change that schema.
2. Since data produced with AvroCompat requires that the consumer import AvroCompat to read it, this design requires the consumer to have knowledge of how the data was produced.

IMO, the Avro compatibility design could be re-worked in a way that’s (a) easy and intuitive for producers to enable, and (b) doesn’t require the data consumer to know about how the upstream was produced. Possibly we should also make Avro-compatible the default way that Magnolify-Parquet writes repeated field fields, so that it’s opt-out, not opt-in, although that’s a potentially big breaking change for users.

clairemcginty:
Magnolify is now on Parquet 0.14, which includes a bugfix for [PARQUET-2425](https://issues.apache.org/jira/browse/PARQUET-2425) -- `AvroSchemaConverter` no longer throws an exception when parsing non-grouped repeated fields (the magnolify-parquet default). This was an early blocker in our efforts to deprecate AvroCompat (see discussion on #766).

This PR:

- Deprecates `AvroCompat`  in favor of (a) making `writeSupport`/`schema` operations Configurable, and introduces a Configuration option specifically for writing grouped arrays
- Updates `readSupport` to automatically detect any grouped arrays in the the Write schema and correct the Read schema as needed.
- Writes Avro schemas to file metadata footer by default


parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -51,7 +51,7 @@ sealed trait ParquetField[T] extends Serializable {
   protected final def nonEmpty(v: T): Boolean = !isEmpty(v)
 
   def write(c: RecordConsumer, v: T)(cm: CaseMapper): Unit
-  def newConverter: TypeConverter[T]
+  def newConverter(writerSchema: Type): TypeConverter[T]
```
clairemcginty, line 54: Alternately, we could create an overloaded method like:

```scala
def newConverter(): TypeConverter = newConverter(false)
def newConverter(avroCompat: Boolean): TypeConverter = ...
```

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -339,15 +343,16 @@ object ParquetField {
...
+        val buffered = t
+          .newConverter(writerSchema)
           .asInstanceOf[TypeConverter.Buffered[T]]
           .withRepetition(Repetition.REPEATED)
         val arrayConverter = new TypeConverter.Delegate[T, C[T]](buffered) {
           override def get: C[T] = inner.get(fc.fromSpecific)
         }
 
-        if (hasAvroArray) {
+        if (Schema.hasGroupedArray(writerSchema)) {
```
clairemcginty, line 355: I guess this would throw an error if the schema mixed grouped and non-grouped array types since it checks for the presence of a grouped array in the entire schema, rather than for the specific field.... but we don't have access to `CaseMapper` here so we couldn't easily isolate the specific field schema. The original approach with the `AvroCompat` import was also all-or-nothing so this shouldn't functionally be a change in behavior

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
RustedBones: This will allow forward binary compatibility. Otherwise we should move this to the v0.8 base branch

clairemcginty: hmm. forward compat would be nice, but long-term I'd like to be able to phase out the whole AvroCompat option from Magnolify and just write grouped arrays by default... so I'd rather not encode it further into the Magnolify API (`def newConverter(avroCompat: Boolean)`). So I think we can keep this as is and I'll rebase onto v0.8 branch

clairemcginty: although now that I think about it more, the param here should be readSchema/requestedSchema, not `writerSchema`

build.sbt
```diff
@@ -693,6 +693,7 @@ lazy val tools = project
       "com.google.apis" % "google-api-services-bigquery" % bigqueryVersion,
       "org.apache.avro" % "avro" % avroVersion % Provided,
       "org.apache.parquet" % "parquet-hadoop" % parquetVersion,
+      "org.apache.hadoop" % "hadoop-common" % hadoopVersion,
```
clairemcginty, line 696: I'm on the fence about relying so heavily on the hadoop `Configuration` class, from the hadoop-common artifact. Parquet is trying to move away from `Configuration` and onto their own [ParquetConfiguration](https://github.com/apache/parquet-java/blob/master/parquet-common/src/main/java/org/apache/parquet/conf/ParquetConfiguration.java) class, which we could use instead. However, it might be confusing for Scio users since Scio is heavily dependent on `Configuration` and we don't have immediate plans to offboard from it

parquet/src/main/scala/magnolify/parquet/ParquetType.scala
```diff
@@ -175,26 +178,40 @@ object ParquetType {
...
       if (parquetType == null) {
         parquetType = SerializationUtils.fromBase64[ParquetType[T]](configuration.get(WriteTypeKey))
       }
 
-      val schema = Schema.message(parquetType.schema)
+      val schema = Schema.message(parquetType.schema(configuration))
       val metadata = new java.util.HashMap[String, String]()
-      if (parquetType.avroCompat) {
-        // This overrides `WriteSupport#getName`
-        metadata.put(ParquetWriter.OBJECT_MODEL_NAME_PROP, "avro")
```
clairemcginty, line 188: I did drop the behavior of writing `writer.model.name: avro` if AvroCompat is enabled -- I don't think it makes sense, it should still be Magnolify. I can't think of any reason why this would impact downstream readers -- model name shouldn't matter at all when comparing schema compatibility across files. lf anyone can think of a good reason why this change is breaking let me know...

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -327,27 +350,33 @@ object ParquetField {
         }
       }
 
-      override protected val isGroup: Boolean = hasAvroArray
+      override protected def isGroup(conf: Configuration): Boolean =
+        groupAvroArrays || conf.getBoolean(
+          MagnolifyParquetProperties.WriteGroupedArrays,
+          MagnolifyParquetProperties.WriteGroupedArraysDefault
+        )
```
RustedBones, line 357: I'm afraid this check in `conf` is very costly and has a significant impact on performance. memorization can help, but IMHO we should give the `configuration` when creating the `ParquetType` (same place we actually capture the deprecated `ParquetArray`), so we can compute this only once. WDYT ?

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -309,14 +329,17 @@ object ParquetField {
...
   ): ParquetField[C[T]] = {
     new ParquetField[C[T]] {
-      override val hasAvroArray: Boolean = pa match {
+      // Legacy compat with Magnolify <= 0.7; future versions will remove AvroCompat in favor of
+      // Configuration-based approach
+      @nowarn("cat=deprecation")
+      val groupAvroArrays: Boolean = pa match {
         case ParquetArray.default               => false
         case ParquetArray.AvroCompat.avroCompat => true
       }
```
RustedBones, line 338: This was probably not ideal. I think we should be able to re-use any `ParquetField` and decide on the behavior when instantiating the parquet type. Here we now mix behaviors with the config.

jmh/src/test/scala/magnolify/jmh/MagnolifyBench.scala
```diff
@@ -87,6 +92,77 @@ class AvroBench {
...
+}
+
+@BenchmarkMode(Array(Mode.AverageTime))
+@OutputTimeUnit(TimeUnit.NANOSECONDS)
+@State(Scope.Thread)
+class ParquetBench {
+  import MagnolifyBench._
+
+  @Benchmark def parquetWrite(state: ParquetStates.DefaultParquetWriteState): Unit = state.writer.write(nested)
+  @Benchmark def parquetRead(state: ParquetStates.DefaultParquetReadState): Nested = state.reader.read()
```
clairemcginty, line 153: It's hard to capture a "true" read benchmark for Parquet since there's so much happening under the hood here (reading and caching the row group, for example). But at least this can be used to track positive and negative trends

build.sbt
clairemcginty: Actually I might pull this out into a separate PR. will update shortly

clairemcginty:
Ok, using the benchmarks from #1040, this change looks performance-neutral on read/write:

`main` branch benchmark:
```
% sbt "jmh/jmh:run -i 10 -wi 10 -f1 -t .*parquet.*magnolify"
[info] Benchmark                           Mode  Cnt      Score     Error  Units
[info] ParquetBench.parquetReadMagnolify  avgt   10  13695.172 ± 311.972  ns/op
[info] ParquetBench.parquetWriteMagnolify  avgt   10  5527.228 ± 70.377  ns/op
```

`avro-compat-rework` benchmark:
```
% sbt "jmh/jmh:run -i 10 -wi 10 -f1 -t .*parquet.*magnolify"
[info] Benchmark                           Mode  Cnt      Score     Error  Units
[info] ParquetBench.parquetReadMagnolify   avgt   10  13491.778 ± 124.734  ns/op
[info] ParquetBench.parquetWriteMagnolify  avgt   10   4769.038 ±  61.517  ns/op
```

the Configuration check does add some time to WriteContext#init, but that's called once per task only

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -29,44 +29,59 @@ import org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName
...
   def typeDoc: Option[String]
 
-  protected val isGroup: Boolean = false
+  protected def isGroup(groupArrayFields: Boolean): Boolean = false
   protected def isEmpty(v: T): Boolean
   protected final def nonEmpty(v: T): Boolean = !isEmpty(v)
 
-  def write(c: RecordConsumer, v: T)(cm: CaseMapper): Unit
-  def newConverter: TypeConverter[T]
+  def write(c: RecordConsumer, v: T)(cm: CaseMapper, groupArrayFields: Boolean): Unit
```
clairemcginty, line 64: so this solution now works by passing a precomputed `groupArrayFields` flag around to all `write` method.

It works, but I'm thinking that we may want to generalize this from a boolean flag into a `Map[String, _]`-typed configuration field (parsed from the Hadoop `Configuration` object passed to `ParquetType`).

This would give us more flexibility if we need to make any more write options configurable in future. As an example, if a user has any `LocalDate*` fields in their `ParquetType` case class, by default parquet-avro will convert them into a `local-timestamp-{millis, micros}` for Avro 1.11, but `timestamp-{millis, micros}` on Avro 1.8. We might want to make this behavior configurable to preserve schema compatibility across Avro upgrades.

clairemcginty: wdyt @RustedBones ?

parquet/src/test/scala/magnolify/parquet/ParquetTypeSuite.scala
```diff
@@ -192,6 +195,59 @@ class ParquetTypeSuite extends MagnolifySuite {
       assertEquals(inner.getFields.asScala.map(_.getName).toSeq, Seq("INNERFIRST"))
     }
   }
+
+  test(s"AvroCompat") {
```
RustedBones, line 199:
```suggestion
  test("AvroCompat") {
```

RustedBones (commented):
I think we should speak of `list logical-type` instead of `grouped array`.

I'm also wondering if by default we should respect the [spec](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md).

`parquet-avro` does not respect the default, and falls in the [backward compat rules](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#backward-compatibility-rules), case 1.
Do you know if `parquet-avro` is able to read arrays when the default spec is used ?

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
RustedBones: We can use a custom trait containing all desired write setting

clairemcginty:
> `parquet-avro` does not respect the default, and falls in the [backward compat rules](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#backward-compatibility-rules), case 1.
> Do you know if `parquet-avro` is able to read arrays when the default spec is used ?

Unfortunately, [not](https://github.com/apache/parquet-java/blob/apache-parquet-1.14.4/parquet-avro/src/test/java/org/apache/parquet/avro/TestArrayCompatibility.java#L60-L82). It looks like this has just never been implemented. It's something we could try to get into the next Parquet release

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
clairemcginty: updated to use a trait!

RustedBones (approved):

parquet/src/main/scala/magnolify/parquet/ParquetField.scala
kellen: @clairemcginty Did you want to make any change here?

kellen (approved):

kellen (approved):

PR #1003, opened 2024-07-22
[avro] Use SchemaBuilder API

RustedBones:
Move implementation to the SchemaBuilder API.
This removes the cluncky JSON conversion type required for default values

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -338,7 +338,7 @@ jobs:
 
       - name: Test
         env:
-          JAVA_OPTS: '-Davro.version=1.8.2'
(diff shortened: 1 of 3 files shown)


RustedBones:
Avro 1.8 fails due to [bug](https://issues.apache.org/jira/browse/AVRO-1275)

clairemcginty (approved):
much better, especially the JsonProperties defaults 🙏 since it does drop Avro 1.8 support, shall we make a branch/milestone to merge this into?

kellen:
If we adopt this into magnolify and subsequently into scio this means that we would force an upgrade past avro 1.8, correct?

RustedBones:
That's right. When mergeing this, avro 1.8 support will be dropped.

PR #1004, opened 2024-07-22
[avro] Add Either typeclass support

RustedBones:
Support scala Either as a avro type union


avro/src/main/scala/magnolify/avro/AvroType.scala
```diff
@@ -253,6 +253,31 @@ object AvroField {
...
+    fb: AvroField[B]
+  ): AvroField[Either[A, B]] =
+    new Aux[Either[A, B], Any, Any] {
+      override protected def buildSchema(cm: CaseMapper): Schema =
+        Schema.createUnion(fa.schema(cm), fb.schema(cm))
+      // `Either[A, B]` is a `UNION` of `[A, B]` and must default to first type `A`
+      override def makeDefault(d: Either[A, B])(cm: CaseMapper): fa.ToT = {
+        require(d.isLeft, "Either[A, B] can only default to Left[A]")
+        fa.to(d.left.get)(cm): @nowarn
+      }
```
RustedBones, line 267: In scala `Either` is `Right` biased but here I chose `Left` as possible default based on type order definition. WDYT?

PR #1040, opened 2024-09-18
Add benchmarks for magnolify-parquet vs parquet-avro R/W

clairemcginty:
Adds benchmarks for Parquet read/write performance, for both magnolify-parquet and parquet-avro (although we don't own parquet-avro, it's helpful to compare against IMO).

Parquet is a little tricky in that it doesn't have a granular "write/read a single record to/from a file" operation due to its complex file structure/encodings. This benchmark sets up an in-memory page store that can can read or write Parquet "groups", which are Parquet's internal record structure. Read/write is invoked with a record type `T` and a matching `RecordConverter[T]`, which converts either case classes (magnolify-parquet) or Avro records (parquet-avro) into Parquet groups. Thus, what we're benchmarking here is Group-to-record and record-to-Group conversion, which is the core functionality of magnolify-parquet 👍 

Results:

```
[info] Benchmark                           Mode  Cnt      Score     Error  Units
[info] ParquetAvroBench.parquetRead        avgt   10  14274.174 ± 414.389  ns/op
[info] ParquetAvroBench.parquetWrite       avgt   10  21282.044 ± 409.851  ns/op
[info] ParquetMagnolifyBench.parquetRead   avgt   10  11393.528 ± 211.746  ns/op
[info] ParquetMagnolifyBench.parquetWrite  avgt   10  13282.019 ± 180.685  ns/op
```

### DIFF
--- a/jmh/src/test/scala/magnolify/jmh/MagnolifyBench.scala
+++ b/jmh/src/test/scala/magnolify/jmh/MagnolifyBench.scala
@@ -160,7 +160,8 @@ class ExampleBench {
   private val exampleNested = implicitly[Arbitrary[ExampleNested]].arbitrary(prms, seed).get
   private val example = exampleType.to(exampleNested).build()
   @Benchmark def exampleTo: Example.Builder = exampleType.to(exampleNested)
-  @Benchmark def exampleFrom: ExampleNested = exampleType.from(example.getFeatures.getFeatureMap.asScala.toMap)
+  @Benchmark def exampleFrom: ExampleNested =
+    exampleType.from(example.getFeatures.getFeatureMap.asScala.toMap)
 }
 
 @BenchmarkMode(Array(Mode.AverageTime))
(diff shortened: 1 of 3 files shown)


jmh/src/test/scala/magnolify/jmh/MagnolifyBench.scala
```diff
@@ -157,7 +160,152 @@ class ExampleBench {
   private val exampleNested = implicitly[Arbitrary[ExampleNested]].arbitrary(prms, seed).get
   private val example = exampleType.to(exampleNested).build()
   @Benchmark def exampleTo: Example.Builder = exampleType.to(exampleNested)
-  @Benchmark def exampleFrom: ExampleNested = exampleType.from(example)
+  @Benchmark def exampleFrom: ExampleNested =
+    exampleType.from(example.getFeatures.getFeatureMap.asScala.toMap)
```
clairemcginty, line 164: this needed to be updated from the ExampleType rework

jmh/src/test/scala/magnolify/jmh/ParquetInMemoryPageStore.scala
```diff
@@ -0,0 +1,130 @@
...
+import org.apache.parquet.column.page._
+import org.apache.parquet.column.statistics._
+
+import scala.collection.mutable
+
+/**
+ * An in-memory Parquet page store modeled after parquet-java's MemPageStore, used to benchmark
+ * ParquetType conversion between Parquet Groups and Scala case classes
+ */
+class ParquetInMemoryPageStore(rowCount: Long) extends PageReadStore with PageWriteStore {
```
clairemcginty, line 30: This class is heavily based on this parquet-java package, which sadly is not a part of any artifact: https://github.com/apache/parquet-java/tree/master/parquet-column/src/test/java/org/apache/parquet/column/page/mem

jmh/src/test/scala/magnolify/jmh/MagnolifyBench.scala
RustedBones: IMHO we should bench the `apply` methods

clairemcginty: Yeah, that would be good! I'll do that in a followup PR

RustedBones (approved):

PR #1169, opened 2025-05-01
(WIP) Support 3 level list encoding

clairemcginty:
Adds support for [3-level list encoding](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists) for read or write.

This PR has some breaking changes in the API signature but I think overall it's worth the pain. It does maintain compatibility with the default `AvroCompat` import behavior. (In fact most of the code complexity here is not from the read/write logic for lists, but rather with keeping as much backwards compatibility as possible...)

I also decided to remove the previous read behavior where we would try to automatically infer what kind of list encoding was used in the file and convert to the correct schema under the hood. I think that's just too hard to do accurately for 3 different list encoding types, lol. Now the user has to explicitly configure what type of list encoding they're reading via a `MagnolifyParquetProperties` or implicit `Configuration`. (This matches what parquet-avro does for array types, where you have to configure an option `parquet.avro.write-old-list-structure -> [true, false]`.) Maybe I could work on providing a more helpful error message for this though

todo: update docs

### DIFF
--- a/parquet/src/main/scala/magnolify/parquet/MagnolifyParquetProperties.scala
+++ b/parquet/src/main/scala/magnolify/parquet/MagnolifyParquetProperties.scala
@@ -16,11 +16,13 @@
 
 package magnolify.parquet
 
+import magnolify.parquet.ArrayEncoding.Ungrouped
+
 trait MagnolifyParquetProperties extends Serializable {
-  def WriteAvroCompatibleArrays: Boolean = false
+  def writeArrayEncoding: ArrayEncoding = Ungrouped
   def writeAvroSchemaToMetadata: Boolean = true
 
-  private[parquet] final def schemaUniquenessKey: Int = WriteAvroCompatibleArrays.hashCode()
+  private[parquet] final def schemaUniquenessKey: Int = writeArrayEncoding.hashCode()
 }
 
 /**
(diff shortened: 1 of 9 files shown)


parquet/src/main/scala/magnolify/parquet/ParquetField.scala
```diff
@@ -334,81 +336,122 @@ object ParquetField {
...
+            v.foreach(t.writeGroup(c, _)(cm, properties))
+            c.endField(AvroArrayField, 0)
+          case ArrayEncoding.ThreeLevel =>
+            c.startField(AvroListField, 0)
+            c.startGroup()
+
+            v.foreach { elem =>
+              c.startField("element", 0)
+              t.write(c, elem)(cm, properties)
+              c.endField("element", 0)
```
charles-tan, line 394: I see some warnings for test coverage for this new code block. Do we test anywhere the write path?

clairemcginty: good catch! that's on my to do list haha...

parquet/src/main/scala/magnolify/parquet/MagnolifyParquetProperties.scala
```diff
@@ -30,6 +30,27 @@ trait MagnolifyParquetProperties extends Serializable {
...
+
   val WriteAvroSchemaToMetadata: String = "magnolify.parquet.write-avro-schema"
 }
+
+sealed trait ArrayEncoding
+
+object ArrayEncoding {
+  case object Ungrouped extends ArrayEncoding
+  case object OldArrayEncoding extends ArrayEncoding
+  case object NewListEncoding extends ArrayEncoding
```
clairemcginty, line 46: Not sure if I 100% like the naming conventions on these. In parquet-avro the equivalent property is called [parquet.avro.write-old-list-structure](https://github.com/apache/parquet-java/blob/apache-parquet-1.15.2/parquet-avro/src/main/java/org/apache/parquet/avro/AvroWriteSupport.java#L69)

charles-tan (approved):
Some suggestions around documentation and naming, but lgtm.

docs/parquet.md
```diff
@@ -55,115 +55,144 @@ val pfDecimalBinary = ParquetField.decimalBinary(20, 0)
...
-When Parquet-Avro Compatibility Mode is enabled, magnolify-parquet will interpret repeated fields using the same 2-level list structure that parquet-avro uses.
-In addition, Parquet file writes will include an extra metadata key, `parquet.avro.schema`, to the file footer, containing the converted, String-serialized Avro schema.
+Magnolify-parquet offers support for all three of these encodings:
 
-#### Enabling Compatibility Mode on Magnolify < 0.8
+|                 | Ungrouped   | Old List Format | New List Format | Configuration                                                            |
+|-----------------|-------------|-----------------|-----------------|--------------------------------------------------------------------------|
+| Magnolify < 0.8 | x (Default) | x               |                 | AvroCompat import                                                        |
+| Magnolify 0.8   | x (Default) | x               |                 | AvroCompat import (Deprecated); `magnolify.parquet.write-grouped-arrays` |
+| Magnolify > 0.8 | x (Default) | x               | x               | AvroCompat import (Deprecated); `magnolify.parquet.write-array-encoding` |
```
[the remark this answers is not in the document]
clairemcginty: would love to, but I think it's too dangerous of a change to make :( Since not all Parquet formats can read different list types interchangeably, we could cause some chaos by by changing the write schema under the hood.

PR #1217, opened 2025-09-15
(fix #1153) Support xmap in magnolify-bigquery

### LINKED ISSUE #1153: Support xmap in TableRowType
Scio's BigQueryType supports type overrides via [OverrideTypeProvider](https://spotify.github.io/scio/internals/OverrideTypeProvider.html). This is used heavily by libs like Elitzur to do field-level validation at runtime. However, there is currently not a migration path for case classes with overriden types to migrate to magnolify-bigquery.

We could support this by unsealing the `TableRowType` trait/adding some kind of `xmap` method so that users can define their own type overrides in Magnolify.

clairemcginty:


bigquery/src/main/scala/magnolify/bigquery/TableRowType.scala
```diff
@@ -177,6 +177,15 @@ object TableRowField {
 
   def apply[T](implicit f: TableRowField[T]): TableRowField[T] = f
 
+  def xmap[U, T](f: U => T, t: T => U)(implicit trf: TableRowField[U]): TableRowField[T] =
+    new Aux[T, trf.FromT, Any] {
```
kellen, line 181: why not `trf.ToT` instead of `Any`?

kellen (approved):

PR #1340, opened 2026-06-23
Tighten schema compatibility checks for repeated types

clairemcginty:
Adds new failure cases to Schema#checkCompatibility:

- When reader and writer [list encoding types](https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists) mismatch (`list` vs `array`)
- When reader requests a required list field that is not present in writer schema

### DIFF
--- a/parquet/src/main/scala/magnolify/parquet/Schema.scala
+++ b/parquet/src/main/scala/magnolify/parquet/Schema.scala
@@ -99,6 +99,9 @@ private object Schema {
     def listFields(gt: GroupType) =
       s"[${gt.getFields.asScala.map(f => s"${f.getName}: ${f.getRepetition}").mkString(",")}]"
 
+    def isRepeatedType(lta: LogicalTypeAnnotation): Boolean =
+      lta == LogicalTypeAnnotation.listType() || lta == LogicalTypeAnnotation.mapType()
+
     def isRepetitionBackwardCompatible(w: Type, r: Type) =
       (w.getRepetition, r.getRepetition) match {
         case (Repetition.REQUIRED, Repetition.OPTIONAL) => true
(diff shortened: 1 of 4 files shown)


parquet/src/test/scala/magnolify/parquet/SchemaEvolutionSuite.scala
```diff
@@ -390,7 +391,10 @@ class SchemaEvolutionSuite extends MagnolifySuite {
   }
 
   test("Scala V1 => Scala V2") {
-    assertEquals(readScala[User2](scalaBytes1), scala1as2)
+    // V2 adds aliases (REPEATED) which is not present in V1
```
clairemcginty: The added checks make these cases fail, but IMO, I don't think they should have passed in the first case - I can't think of a situation where you would want a required list field missing in the writer schema to get read as `Nil` 🤷‍♀️

parquet/src/main/scala/magnolify/parquet/Schema.scala
```diff
@@ -122,21 +122,17 @@ private object Schema {
...
-                throw new InvalidRecordException(
-                  s"Requested field `${rf.getName}: ${rf.getRepetition}` is not present in written file schema. " +
-                    s"Available fields are: ${listFields(wg)}"
-                )
-              case (true, Repetition.OPTIONAL) =>
+            rf.getRepetition match {
+              case Repetition.OPTIONAL =>
                 logger.warn(
                   s"Requested field `${rf.getName}: ${rf.getRepetition}` is not present in written file schema " +
                     s"and will be evaluated as `Option.empty`. Available fields are: ${listFields(wg)}"
```
kellen: Is Option.empty the same as Nil for list types here? If not, can you update the message?

parquet/src/test/scala/magnolify/parquet/SchemaEvolutionSuite.scala
kellen: Isn't this a forwards compatibility thing? Like we keep telling people not to use `Option[List[]]` so if they want to add a list type it has to be REQUIRED and interpreted as Nil if reading from a previous version

clairemcginty: Trueee maybe we should narrow the scope of this then to only throw if (a) the parent list field name exists in both reader and writer, and (b) a different encoding is used. wdyt

kellen: That sounds better yes

kellen (approved):

PR #1341, opened 2026-06-23
Add Hadoop config option to auto-derive Parquet list encoding from writer schema

clairemcginty:
(branched off https://github.com/spotify/magnolify/pull/1340; once that's merged I'll update the base ref)

Adds a hadoop configuration flag `parquet.type.read.auto-detect-list-encoding` that will instruct ParquetType to auto-detect list encoding based off writer schema.

todo: update benchmarks to check added ms cost per file

### DIFF
--- a/docs/parquet.md
+++ b/docs/parquet.md
@@ -117,6 +117,35 @@ case class RecordWithList(listField: List[String])
 ParquetType[RecordWithList].schema
 ```
 
+### Auto-detecting list encodings
+
+As of Magnolify 0.9.5, passing in the Hadoop configuration option `parquet.type.read.auto-detect-list-encoding: true` at read time will instruct `ParquetType` to automatically infer list encoding from the writer schema,
+obviating the need to specify an encoding manually.
+
+```diff
+val pt = ParquetType[RecordWithList](
+-  new MagnolifyParquetProperties {
+-    override def writeArrayEncoding: ArrayEncoding = ArrayEncoding.ThreeLevelList
+-  }
+)
+
+val reader = pt
+  .readBuilder(...)
++ .withConf(ParquetConfiguration.of("parquet.type.read.auto-detect-list-encoding" -> true)
+  .build()
+```
+
+You can also set this property in a `core-site.xml` file in `src/main/resources/` to be automatically applied to every read in your project:
+
+```xml
+<configuration>
+  <property>
+    <name>parquet.type.read.auto-detect-list-encoding</name>
+    <value>true</value>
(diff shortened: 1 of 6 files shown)


parquet/src/main/scala/magnolify/parquet/ParquetType.scala
```diff
@@ -167,6 +169,9 @@ object ParquetType {
 
         override private[parquet] def properties: MagnolifyParquetProperties =
           propertiesWithAvroImportCompat
+        override private[parquet] def updateProperties(
```
clairemcginty, line 172: this feels like a very non-idiomatic way to achieve this in Scala, but couldn't think of a better option that wouldn't break API compat

kellen: Maybe a private `apply` would seem nicer?

kellen (approved):

docs/parquet.md
```diff
@@ -117,6 +117,35 @@ case class RecordWithList(listField: List[String])
...
+```diff
+val pt = ParquetType[RecordWithList](
+-  new MagnolifyParquetProperties {
+-    override def writeArrayEncoding: ArrayEncoding = ArrayEncoding.ThreeLevelList
+-  }
+)
+
+val reader = pt
+  .readBuilder(...)
++ .withConf(ParquetConfiguration.of("parquet.type.read.auto-detect-list-encoding" -> true)
```
regadas, line 134: nit: missing `)`
```suggestion
+ .withConf(ParquetConfiguration.of("parquet.type.read.auto-detect-list-encoding" -> true))
```

parquet/src/main/scala/magnolify/parquet/Schema.scala
```diff
@@ -95,13 +95,50 @@ private object Schema {
...
+
+    if (encodings.size > 1) {
+      throw new InvalidRecordException(
+        s"Multiple list encodings detected in writer schema: ${encodings.mkString(", ")}. " +
+          "Cannot auto-detect list encoding."
+      )
+    }
+
+    encodings.headOption
+  }
```
regadas, line 136: hmm maybe I'm wrong but the current 2-level-array test fixtures use a group element, so we're missing a primitive-element unit test + roundtrip? If so I think this path might be broken for 2-level array encoding over primitive elements ... throwing "Multiple list encodings".

regadas (changes requested):
one thing worth confirming before moving ahead.

PR #1350, opened 2026-07-10
(magnolify-beam) RowType#from should look up field values by name, not index

clairemcginty:
IcebergIO with beam `Row` isn't like Avro in the sense that you pass it a reader schema ahead of time and it will decode Iceberg records into `Rows` with where the field index is always guaranteed to match your Magnolify#RowType schema. (From what I can tell it will always match the order of the Iceberg table schema itself; passing the `keep` config option with rows in the different orders doesn't affect the returned Row schema at all). Therefore we should look up fields by name, not index.

~Won't undraft until I run the benchmarks, it will probably be a bit slower :( this is potentially something we could look into fixing on the Beam side~

Update 7/14: with benchmarks, this does slow down RowType#rowFrom quite a lot.

Baseline (`main` branch):
```
[info] Benchmark         Mode  Cnt      Score     Error  Units
[info] RowBench.rowFrom  avgt   10    984.269 ±   7.242  ns/op
[info] RowBench.rowTo    avgt   10  26021.345 ± 141.150  ns/op
```

This PR branch:
```
[info] Benchmark         Mode  Cnt      Score     Error  Units
[info] RowBench.rowFrom  avgt   10   1620.904 ±   1.539  ns/op
[info] RowBench.rowTo    avgt   10  25987.984 ± 930.496  ns/op
```

Update 7/15: pushed https://github.com/spotify/magnolify/commit/248c272d5fd8cffa1fa282cfb25069a8705c7819 which caches field indices after encountering them in the first `from` invocation, which significantly improves RowType#from performance:

```
[info] Benchmark         Mode  Cnt     Score   Error  Units
[info] RowBench.rowFrom  avgt   10  1053.535 ± 9.473  ns/op
```

I think this is close enough to baseline that i'll undraft the PR.

### DIFF
--- a/beam/src/main/scala/magnolify/beam/RowType.scala
+++ b/beam/src/main/scala/magnolify/beam/RowType.scala
@@ -171,7 +171,9 @@ object RowField {
             .build()
 
         override def from(v: Row)(cm: CaseMapper): T =
-          caseClass.construct(p => p.typeclass.fromAny(v.getValue[Any](p.index))(cm))
+          caseClass.construct { p =>
+            p.typeclass.fromAny(v.getValue[Any](cm.map(p.label)))(cm)
+          }
 
         override def to(v: T)(cm: CaseMapper): Row = {
           val values = caseClass.parameters.map { p =>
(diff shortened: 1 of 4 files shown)


regadas (approved):

kellen (approved):
yikes!

